PR #1, opened 2016-08-13
Fixed RHEL-07-010490, RHEL-07-020200, RHEL-07-021600, RHEL-07-021610,…

zer0glitch:
… RHEL-07-040010, RHEL-07-040300, RHEL-07-040320


defaults/main.yml
```diff
@@ -48,3 +48,15 @@ rhel7stig_boot_password_config:
     - regexp: ^password_pbkdf2 root
       line: password_pbkdf2 root {{ rhel7stig_bootloader_password | grub2_hash(salt='KeokpkECTJeoDhEA5XtiLQ') }}
 
+rhel7stig_dns_servers:
+  - 8.8.8.8
```
samdoran: Please indicate what vuln these variables are for and brief description of their puprpose.

Please indent four spaces in accordance with the [guidelines](https://github.com/ansible/ansible-lockdown/blob/master/CONTRIBUTING.md).

library/rhel7_020840.py
```diff
@@ -0,0 +1,71 @@
...
+    try:
+      changed = init_permissions(module.params.get('check'))
+    except Exception as ex:
+        module.fail_json(msg=str(ex))
+
+    res_args = dict(
+        changed = changed, msg = "OK"
+    )
+
+    module.exit_json(**res_args)
```
samdoran: It might makes sense to return the ouput to return its output inside ansible_facts, which is the [recommended way](http://docs.ansible.com/ansible/developing_modules.html#module-provided-facts) to create custom modules that gather facts.

tasks/audit-cat1.yml
```diff
@@ -103,6 +103,8 @@
       - RHEL-07-040331
       - always
 
+- debug: var={{ rhel_07_040331_audit }}
```
samdoran: Not sure if this is supposed to be here. If it is, please see the guidelines for proper task naming and do not Ansible shorthand syntax.

tasks/audit-cat2.yml
```diff
@@ -77,3 +77,88 @@
       - RHEL-07-040640
       - always
       - ssh
+
+- name: "MEDIUM | RHEL-07-020840 | PATCH | All local initialization files for interactive users must be owned by the home directory user or root."
```
samdoran: This is and audit task, not a patch task.

tasks/audit-cat2.yml
```diff
@@ -77,3 +77,88 @@
...
+  rhel7_020840: check=True
+  failed_when: no
+  changed_when: no
+  ignore_errors: yes
+  register: rhel_07_020840_audit
+  tags:
+      - cat2
+      - medium
+      - patch
```
samdoran: This is and audit task, not a patch task.

There are several task below in the same boat. `s/audit/patch/gi`

tasks/audit-cat3.yml
```diff
@@ -1,6 +1,152 @@
-- name: "Place holder for Cat III Audits"
-  command: "true"
+---
```
samdoran: The `---` is not needed and can be omitted.

tasks/audit-cat3.yml
```diff
@@ -1,6 +1,152 @@
...
+  with_items:
+    - games
+    - ftp
+  tags:
+    - cat3
+    - low
+    - audit
+    - RHEL-07-010490
+
+- fail: msg="Unecessary default account present."
```
samdoran: Please do not use the Ansible shorthand syntax. Also, a `name` field is required per the contribution guidelines.

```
- name: "LOW | RHE-XX-XXXXXX | PATCH | Description"
  fail:
      msg: "Unnecessary default account present."
```

samdoran: But more importantly, we don't want to insert things that cause the role to fail. It needs to run to completion by design, and report via a template output at the end (currently unbuilt) the status of findings.

Failing immediately makes sense when running from the command line, but is a very poor experience when run from Tower, or if the job template is being run from other service via an API call to Tower.

tasks/audit-cat2.yml
zer0glitch: Should we be changing the permissions on the system?  So if .bashrc in /root is owned by apache, should we correct or just show the error?

tasks/audit-cat1.yml
zer0glitch: Removed this in latest PR

tasks/audit-cat2.yml
samdoran: Yes, the role should correct the permissions.

Per the fix text from the STIG, the file should be owned by the user whose directory it is in. This will most likely involve the combination of an audit/prelim task that first finds all those files, then the actually patch task using the file module to set the owner of the file based on the username extracted from the found file path.

zer0glitch: The module handles the bulk of the load on this instead of a multi-step or extra long command.  Here is the logic of the init_fix.py module

1.  find all users that have the ability to login
2. Get the home directories for these users
3. Pull all files/dirs that start with .
4. Operate on only files
5.  Call the appropriate function fix(user, group, permissions), validate paths.  When operating in check/audit mode it will return any files that do not pass, in fix mode it will resolve any issues that it can.

skip0321:
Is there an ETA on this merge?

tasks/audit-cat1.yml
shepdelacreme: This doesn't appear to have been removed.

tasks/audit-cat2.yml
shepdelacreme: Tasks in this file are labeled PATCH instead of AUDIT. Should be updated to all be AUDIT, including the tags.

tasks/audit-cat2.yml
```diff
@@ -77,3 +77,88 @@
...
+      - RHEL-07-020840
+
+- name: "MEDIUM | RHEL-07-020850 | PATCH | Local initialization files for local interactive users must be group-owned by the users primary group or root."
+  init_files: check=True group=True
```
shepdelacreme: Per style guide: 
```
init_files: check=True group=True
```
Should be:
```
init_files:
    check: True
    group: True
```

tasks/audit-cat2.yml
```diff
@@ -77,3 +77,88 @@
...
+  ignore_errors: yes
+  register: rhel_07_020850_audit
+  tags:
+      - cat2
+      - medium
+      - patch
+      - RHEL-07-020850
+
+- name: "MEDIUM | RHEL-07-020860 | PATCH | All local initialization files must have mode 0740 or less permissive."
+  init_files: check=True permissions=True
```
shepdelacreme: Per style guide: 
```
init_files: check=True permissions=True
```
Should be:
```
init_files:
    check: True
    permissions: True
```

tasks/audit-cat2.yml
```diff
@@ -77,3 +77,88 @@
...
+  ignore_errors: yes
+  register: rhel_07_020860_audit
+  tags:
+      - cat2
+      - medium
+      - patch
+      - RHEL-07-020860
+
+- name: "MEDIUM | RHEL-07-020870 | PATCH | All local interactive user initialization files executable search paths must contain only absolute paths."
+  init_files: check=True paths=True
```
shepdelacreme: Per style guide: 
```
init_files: check=True paths=True
```
Should be:
```
init_files:
    check: True
    paths: True
```

tasks/fix-cat2.yml
```diff
@@ -466,31 +466,31 @@
       - RHEL-07-020700
 
 - name: "MEDIUM | RHEL-07-020840 | PATCH | All local initialization files for interactive users must be owned by the home directory user or root."
-  command: "true"
+  init_files: check=False user=True
```
shepdelacreme: Per style guide: 
```
init_files: check=False user=True
```
Should be:
```
init_files:
    check: False
    user: True
```

tasks/fix-cat2.yml
```diff
@@ -466,31 +466,31 @@
...
+  init_files: check=False permissions=True
   tags:
       - cat2
       - medium
       - patch
       - RHEL-07-020860
 
 - name: "MEDIUM | RHEL-07-020870 | PATCH | All local interactive user initialization files executable search paths must contain only absolute paths."
-  command: "true"
+  init_files: check=True paths=True
```
shepdelacreme: Per style guide: 
```
init_files: check=False paths=True
```
Should be:
```
init_files:
    check: False
    paths: True
```

tasks/fix-cat3.yml
```diff
@@ -1,13 +1,17 @@
...
-      - patch
-      - RHEL-07-010490
+    - cat3
+    - low
+    - patch
+    - RHEL-07-010490
 
 - name: "LOW | RHEL-07-020200 | PATCH | The operating system must remove all software components after updated versions have been installed."
-  command: "true"
+  lineinfile: dest=/etc/yum.conf regexp=^clean_requirements_on_remove line=clean_requirements_on_remove=1
```
shepdelacreme: See my other style guide comments. This needs to be in YAML literal format.

tasks/fix-cat3.yml
```diff
@@ -22,75 +26,113 @@
...
       - patch
       - RHEL-07-021270
 
-- name: "LOW | RHEL-07-021600 | PATCH | The file integrity tool must be configured to verify Access Control Lists (ACLs)."
-  command: "true"
+- name: |
+
+      "LOW | RHEL-07-021600 | PATCH | The file integrity tool must be configured to verify Access Control Lists (ACLs). | AIDE INSTALL" 
+      "LOW | RHEL-07-021610 | PATCH | The file integrity tool must be configured to verify extended attributes. | AIDE INSTALL"
+  yum: name=aide state=installed
```
shepdelacreme: See my other style guide comments. This needs to be in YAML literal format.

tasks/fix-cat3.yml
```diff
@@ -22,75 +26,113 @@
...
       - patch
       - RHEL-07-021600
 
-- name: "LOW | RHEL-07-021610 | PATCH | The file integrity tool must be configured to verify extended attributes."
-  command: "true"
+- name:  |
+
+      "LOW | RHEL-07-021600 | PATCH | The file integrity tool must be configured to verify Access Control Lists (ACLs). | AIDE CONF" 
+      "LOW | RHEL-07-021610 | PATCH | The file integrity tool must be configured to verify extended attributes. | AIDE CONF"
+  template: dest=/etc/aide.conf src=aide.conf.j2
```
shepdelacreme: See my other style guide comments. This needs to be in YAML literal format.

tasks/fix-cat3.yml
```diff
@@ -22,75 +26,113 @@
...
+      - cat3
+      - low
+      - patch
+      - RHEL-07-021610
+
+- name:  |
+
+      "LOW | RHEL-07-021600 | PATCH | The file integrity tool must be configured to verify Access Control Lists (ACLs). | AIDE DEFAULT CONF DIR"
+      "LOW | RHEL-07-021610 | PATCH | The file integrity tool must be configured to verify extended attributes. | AIDE DEFAULT CONF DIR"
+  file: dest=/etc/aide.conf.d state=directory
```
shepdelacreme: See my other style guide comments. This needs to be in YAML literal format.

tasks/fix-cat3.yml
```diff
@@ -22,75 +26,113 @@
...
+      - cat3
+      - low
+      - patch
+      - RHEL-07-021610
+
+- name:  |
+
+      "LOW | RHEL-07-021600 | PATCH | The file integrity tool must be configured to verify Access Control Lists (ACLs). | AIDE DEFAULT CONF"
+      "LOW | RHEL-07-021610 | PATCH | The file integrity tool must be configured to verify extended attributes. | AIDE DEFAULT CONF"
+  template: dest=/etc/aide.conf.d/default.aide src=default.aide.j2
```
shepdelacreme: See my other style guide comments. This needs to be in YAML literal format.

tasks/fix-cat3.yml
```diff
@@ -22,75 +26,113 @@
...
   tags:
       - cat3
       - low
       - patch
       - RHEL-07-021610
 
 - name: "LOW | RHEL-07-040010 | PATCH | The operating system must limit the number of concurrent sessions to 10 for all accounts and/or account types."
-  command: "true"
+  lineinfile: dest='/etc/security/limits.conf' regexp="^\* hard maxlogins" line="* hard maxlogins {{ rhel7stig_num_max_logins }}"
```
shepdelacreme: See my other style guide comments. This needs to be in YAML literal format.

tasks/fix-cat3.yml
```diff
@@ -22,75 +26,113 @@
...
   tags:
       - cat3
       - low
       - patch
       - RHEL-07-040010
 
 - name: "LOW | RHEL-07-040300 | PATCH | The system must display the date and time of the last successful account logon upon logon."
-  command: "true"
+  lineinfile: dest=/etc/pam.d/postlogin regexp="^session required pam_lastlog.so" line="session required pam_lastlog.so showfailed"
```
shepdelacreme: See my other style guide comments. This needs to be in YAML literal format.

tasks/fix-cat3.yml
```diff
@@ -22,75 +26,113 @@
...
+  shell: grep hosts /etc/nsswitch.conf | grep dns | grep -v '^#' | wc -l
+  register: nsswitch_check
   tags:
       - cat3
       - low
       - patch
       - RHEL-07-040320
 
+- name: "LOW | RHEL-07-040320 | PATCH | For systems using DNS resolution, at least two name servers must be configured. Update if only 1 server"
+  lineinfile: dest=/etc/resolv.conf regexp="^nameserver {{ item }}" line="nameserver {{ item }}"
```
shepdelacreme: See my other style guide comments. This needs to be in YAML literal format.

Issue #9, opened 2017-02-23
running fix_cat2.yml causes centos yum to fail


kbroughton:
I'm guessing it has dis-trusted centos repos or placed a restriction on certs too high for standard servers.  I would expect that the playbook work on centos boxes.

"It was impossible to connect to the CentOS servers.
This could mean a connectivity issue in your environment, such as the requirement to configure a proxy,
or a transparent proxy that tampers with TLS security, or an incorrect system clock.
Please collect information about the specific failure that occurs in your environment,
using the instructions in: https://access.redhat.com/solutions/1527033 and create a bug on https://bugs.centos.org/"

The last successful task and first failure are

TASK [RHEL7-STIG : MEDIUM | RHEL-07-040250 | PATCH | The operating system must protect against or limit the effects of Denial of Service (DoS) attacks by validating the operating system is implementing rate-limiting measures on impacted network interfaces.] ***
Thursday 23 February 2017  10:31:35 -0600 (0:00:01.433)       0:11:13.850 *****
ok: [34.*.*.14]

TASK [RHEL7-STIG : MEDIUM | RHEL-07-040260 | PATCH | All networked systems must have SSH installed.] ***
Thursday 23 February 2017  10:31:38 -0600 (0:00:02.325)       0:11:16.175 *****
[...]

PR #10, opened 2017-02-23
don't apply repo_gpgcheck or local_gpgcheck for centos

kbroughton:
fixes issue #9


shepdelacreme:
See explanation for closure in issue #9

PR #14, opened 2017-05-23
Update the numbering to the release from the draft

dsmorse:
I also implemented a few more lockdowns based on our draft cent7-stig


samdoran:
Great works! Thank you!

PR #16, opened 2017-06-08
Updates to the STIG role

shepdelacreme:
Major changes:

* Finished updating task numbering to V1R1 STIG content.
* Removed audit tasks
* Added variables for each STIG ID so each item can be enabled/disabled via the corresponding variable
* Made some style changes and cleaned up some items that were not working on CentOS 7
* Started to organize tasks into related blocks to apply conditionals and tags to the entire block

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 *.pyc
-
 *.retry
+Vagrantfile
+.vagrant
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -24,7 +24,8 @@ rhel_07_020070: true
 rhel_07_020210: true
 rhel_07_020220: true
 rhel_07_020230: true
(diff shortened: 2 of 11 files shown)


shepdelacreme:
Closed and reopened to trigger CI

dsmorse:
This is great work.. thanks!

fxfitz:
@shepdelacreme How come the audit tasks were commented out?

samdoran:
We decided to remove audit tasks as they caused a good deal of extra tasks not related to doing real remediation. Most of the commands were `shell` or `command` tasks that gathered data but did not make use of it, increasing the runtime and complexity of the role.

Certain audit tasks will be remain if they are needed to patch tasks, but it's no longer a blanket pattern we're implementing.

PR #19, opened 2017-06-15
Update the labels on the AUDIT tasks and set them to run in checkmode

dsmorse:
Also added two new implementations and fixed RHEL-07-010290 so that it would not break the symlink and create /etc/pam.d/system-auth and /etc/pam.d/system-auth-ac as two files


dsmorse:
Updated based on the travis.CI build idempotence failure.  CI build passes on the source branch (now), not sure why it didn't trigger again here.

PR #20, opened 2017-06-20
Firewall Service - Ability to have iptables

fxfitz:
With this PR, one will be able to configure whether or not to use `firewalld` or `iptables` (our systems use iptables, so we need this).


dsmorse:
LGTM

fxfitz:
Any way we could get this merged and released? 🤞 😄

Issue #31, opened 2017-09-20
All modules using validate must use an absolute path to the executable - i.e. sshd tasks


average-joe:
While running CAT1 patches, certain ssh tasks keep failing.  Specifically, **RHEL-07-010300** & **RHEL-07-040390**, however **RHEL-07-040710** does complete.  

These are the results of running each STIG:

**RHEL-07-010300**
```
TASK [MindPointGroup.RHEL7-STIG : HIGH | RHEL-07-010300 | PATCH | The SSH daemon must not allow authentication using an empty password.] 
fatal: [lab-system]: FAILED! => {"changed": false, "cmd": "sshd -tf /tmp/tmppYXE_o", "failed": true, "msg": "[Errno 2] No such file or directory", "rc": 2}
```

**RHEL-07-040390**
```
TASK [MindPointGroup.RHEL7-STIG : HIGH | RHEL-07-040390 | PATCH | The SSH daemon must be configured to only use the SSHv2 protocol.] 
fatal: [lab-system]: FAILED! => {"changed": false, "cmd": "sshd -t -f /tmp/tmpR2L_QS", "failed": true, "msg": "[Errno 2] No such file or directory", "rc": 2}
```

**RHEL-07-040710**
```
TASK [MindPointGroup.RHEL7-STIG : HIGH | RHEL-07-040710 | PATCH | Remote X connections for interactive users must be encrypted.] 
ok: [lab-system]
```
The `validate` line appears to be the issue.  Has this been reproduced by anyone?  I've attempted this on multiple machines with the same result.

shepdelacreme:
Can you share your `ansible --version` and how you are executing the role? I'll see if I can replicate. Also is this against a RHEL or CentOS end host?

average-joe:
Here's what I have:

Ansible version
```
ansible 2.3.1.0
python version = 2.7.13 (default, Apr 20 2017, 12:13:37) [GCC 6.3.0]
```
Below is my command line executing the role:
` ansible-playbook playbooks/stigs.yml -l "hostname"`

I'm running on a CentOS 7.

shepdelacreme:
@average-joe Thanks! I'll take a look. Can I assume `playbooks/stigs.yml` is just basic like so?
```yaml

- hosts: all
  become: yes
  become_method: sudo

  roles:
    - MindPointGroup.RHEL7-STIG
```

average-joe:
@shepdelacreme you're correct.

[...]

PR #32, opened 2017-09-21
Validate command must use absolute path

shepdelacreme:
This is a fix for #31


shepdelacreme:
@fxfitz Yes setting the PATH for each task was also suggested to me as a fix. Would there be an instance on CentOS 7 or RHEL 7 where sshd would not be in /usr/sbin?

samdoran:
Just played around with creative ways to find `sshd` in other paths, and it's really not easy to do if the `PATH` is empty. I think we'll just go with hardcoding this path since it should always be the same on CentOS/RHEL7 and address any issues if they come up.

PR #40, opened 2017-09-29
correct-sshd-validation (RHEL-07-010470)

average-joe:
Identified sshd STIG with old `validate` command.  

Correcting `sshd -t -f %s` to `/usr/sbin/sshd -t -f %s`.


fxfitz:
LGTM

PR #46, opened 2017-11-01
Apply one fix for sshd lastlog

surfr326:
The fix for this rule is to modify /etc/pam.d/sshd or /etc/ssh/sshd_config. When both are modified I observed duplicate lastlog lines being printed.


shepdelacreme:
LGTM thanks!

PR #49, opened 2017-11-07
CAT2 and CAT3 task updates and clean up

shepdelacreme:
This is a work in progress branch for addition of CAT2 and CAT3 tasks as well as some clean up items and rework.

### DIFF
--- a/tasks/fix-cat3.yml
+++ b/tasks/fix-cat3.yml
@@ -12,15 +12,24 @@
       - RHEL-07-020200
 
 - name: "LOW | RHEL-07-020300 | PATCH | All Group Identifiers (GIDs) referenced in the /etc/passwd file must be defined in the /etc/group file."
-  command: "true"
-  changed_when: no
+  block:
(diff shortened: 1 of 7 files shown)


shepdelacreme:
merging these into devel

tasks/fix-cat3.yml
```diff
@@ -12,59 +12,82 @@
...
 
 - name: "LOW | RHEL-07-021310 | PATCH | A separate file system must be used for user home directories (such as /home or an equivalent)."
-  command: "true"
-  changed_when: no
-  when: rhel_07_021310
+  debug:
+      msg: "WARNING: /home is not mounted on a separate partition"
+  when:
+    - rhel_07_021310
+    - ansible_mounts | selectattr('mount', 'match', '^/home$') | list | length == 0
```
jamescassell: This variable is used but never created/defined anywhere.

jamescassell: okay, turns out that it was because I didn't have 'hardware' in my gather_subset in ansible.cfg.

PR #53, opened 2017-12-15
Rhel 07 040530

notkhris:
Good evening all.
I took a crack at resolving cat III issue RHEL-07-040530.
The implementation appears to be safe enough to run multiple times against the same system without adverse effects.

Please review and let me know if you have any concerns or changes I should make.

### DIFF
--- a/tasks/fix-cat3.yml
+++ b/tasks/fix-cat3.yml
@@ -133,15 +133,12 @@
     block: |
         ### RHEL-07-040530 ###
         session required pam_lastlog.so showfailed
-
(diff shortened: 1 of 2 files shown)


notkhris:
I tested this on both rhel7.4 & centos7.4 with no issues regarding /etc/pam.d/postlogin-ac.
Have you guys stopped authconfig from managing the pam files? Should I switch it to change '/etc/pam.d/postlogin'?

shepdelacreme:
Looks good thanks!

PR #54, opened 2017-12-19
Convert rhel7stig_login_defaults.umask to string value.

notkhris:
While running some tests I noticed the UMASK setting in /etc/login.defs was being set to the number 63 instead of the expected 077.
Looking into it, python treats all numbers that begin with '0' as octal notation; '077' in octal is 63 in decimal.
Quoting this value (i.e., changing it into a string) resolves this issue.


shepdelacreme:
Good catch thank you!

Issue #58, opened 2018-01-10
grub2_hash broken with RHEL7 stock python-passlib


jamescassell:
[rhel7 ~]$ rpm -q python-passlib
python-passlib-1.6.5-2.el7.noarch

This role fails due to the required version of passlib.

This should be a trivial fix.  I've added comments here:
https://github.com/MindPointGroup/RHEL7-STIG/pull/23

shepdelacreme:
@jamescassell I think I've addressed this issue in #59 let me know what you think once it passes CI.

Thanks!

jamescassell:
@shepdelacreme PR #59 looks good to me.

shepdelacreme:
Thanks! Closing this issue.

PR #23, opened 2017-08-01
Cleanup grub2_hash filter

samdoran:
- Remove unneceassry imports
- Report an error istead of a stack trace if passlib is not installed

Fixes #21

PR #59, opened 2018-01-10
grub2_hash support for older (<1.7) passlib versions

shepdelacreme:
Per #58 and the comments on #23 some recent systems still have slightly older versions of passlib installed by default. i.e. RHEL7 still uses python-passlib-1.6.5

Updated the filter plugin to support back to 1.5 but it could probably work with as old as 1.4 if necessary. Made some other minor updates as well:

* updated to support older passlib versions back to 1.5
* updated to use the hash() and using() methods on 1.7+ so that the plugin won't die once encrypt() is fully removed
* removed code to generate random salt as passlib does this by default
* removed unneeded python libraries that were used in random salt generation code
* use kwargs to pass user specified values for salt and rounds to the encrypt() or hash() methods
* set default rounds to 25000 to match the default value passlib 1.6+ uses
* catch ImportError instead of generic Exception on the import try/except

### DIFF
--- a/filter_plugins/filters.py
+++ b/filter_plugins/filters.py
@@ -2,7 +2,7 @@ try:
     import passlib.hash
     HAS_PASSLIB = True
     PASSLIB_VERSION = float(passlib.__version__[:3])
-except Exception as e:
+except ImportError as e:
     HAS_PASSLIB = False
 
 from ansible import errors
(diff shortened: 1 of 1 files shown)


jamescassell:
LGTM, though I haven't tested it yet.

shepdelacreme:
I did some local manual testing with pip installed versions of passlib (1.7.1, 1.6.5, and 1.5.3).

jamescassell:
WORKSFORME.  Tested these changes in my environment.  Solves the issue.

shepdelacreme:
Fixes #58

PR #63, opened 2018-03-14
Fix usb-storage blacklist syntax

### LINKED ISSUE #62: Bad modprobe.d syntax in RHEL-07-020100 and RHEL-07-020101
This task:

https://github.com/MindPointGroup/RHEL7-STIG/blob/df72d02aaa8d20bffd2270a2c61266af23cbb16b/tasks/fix-cat2.yml#L620-L635

Adds `^blacklist usb-storage$` to `/etc/modprobe.d/blacklist.conf`.

This is not valid `modprobe.d` syntax and is probably the result of a copy-paste-error.

lerlacher:
Bad syntax for `modprobe.d` blacklist entry

Fixes #62

### DIFF
--- a/tasks/fix-cat2.yml
+++ b/tasks/fix-cat2.yml
@@ -619,10 +619,10 @@
 
 - name: "MEDIUM | RHEL-07-020100 | PATCH | USB mass storage must be disabled."
   lineinfile:
-    state: present
-    create: yes
-    line: '^blacklist usb-storage$'
     dest: /etc/modprobe.d/blacklist.conf
+    regexp: "^(#)?blacklist usb-storage(\\s|$)"
+    line: 'blacklist usb-storage'
+    create: yes
     owner: root
     group: root
     mode: "0644"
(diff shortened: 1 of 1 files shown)


shepdelacreme:
Hmm yes the syntax on that `lineinfile` task and a few others is not so good. We ran into this same issue in the RHEL7-CIS project and came up with this which is a safe regex that should work correctly in all cases.

```yaml
- name: "MEDIUM | RHEL-07-020100 | PATCH | USB mass storage must be disabled."
  lineinfile:
    dest: /etc/modprobe.d/blacklist.conf
    regexp: "^(#)?blacklist usb-storage(\\s|$)"
    line: "blacklist usb-storage"
    create: yes
    owner: root
    group: root
    mode: "0644"
  when: rhel_07_020100
  tags:
      - cat2
      - medium
      - patch
      - RHEL-07-020100
      - usb_devices
```

Do you mind updating to use the above form of the task?

lerlacher:
How 'bout this?

shepdelacreme:
Looks good thanks!

PR #67, opened 2018-03-16
Adjust ssh key paths to /etc/ssh and valid file patterns

mikerenfro:
As 040410 and 040420 are about ssh host key pairs, it should match the intent and produce fewer side effects if we restrict looking for host keys in /etc/ssh and of the format ssh_host*_key and ssh_host*_key.pub.


shepdelacreme:
Thanks!

Squashing and merging this one to get rid of that first Merge commit. If you use rebase against the upstream branch you won't get those.

PR #69, opened 2018-03-19
Audit fixup

jamescassell:
audit loop fixes:
- audit both 32 and 64-bit system calls
- fix -EACCES error-causing typo
- commands audited migrated to audit_commands.yml
- eliminated need for separate "create" parameter
- reformatted code to match guidelines

Please merge rather than squash.


shepdelacreme:
Thanks for these changes I have limited connectivity until Fri/Sat so I may not get to a full review of them until the end of the week.

shepdelacreme:
Looks good...I'm thinking eventually (i.e. not in this PR) it might be preferable to move those large blocks defining all the various audit checks to a var file and then doing a simple `with_items: rhel_7_auditd_syscall_rules` or something? What do you think?

PR #71, opened 2018-03-20
Removing duplicate placeholder code from RHEL-07-021120.

mikerenfro:
Minor cleanup of old placeholder code for RHEL-07-021120.

### DIFF
--- a/tasks/fix-cat2.yml
+++ b/tasks/fix-cat2.yml
@@ -1055,17 +1055,6 @@
       - RHEL-07-021110
       - cron
 
-- name: "MEDIUM | RHEL-07-021120 | PATCH | If the cron.allow file exists it must be group-owned by root."
-  command: "true"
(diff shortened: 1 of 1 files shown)


jamescassell:
Looks good to me...  The check above probably doesn't need to specify mode, and could probably be split into two separate tasks, per the stated goals of the project, but I don't think it's necessary.

tasks/fix-cat2.yml
```diff
@@ -1047,25 +1047,17 @@
...
-  when: rhel_07_021110
+  when:
+    - rhel_07_021110
+    - rhel_07_021120
   tags:
       - cat2
       - medium
       - patch
       - RHEL-07-021110
+      - RHEL-07-021110
```
shepdelacreme: This should be `- RHEL-07-021120`

PR #72, opened 2018-03-20
Implementing fixes for RHEL-07-020620, RHEL-07-020630, RHEL-07-020640, RHEL-07-020650

mikerenfro:


jamescassell:
I don't see a need to put the task into a separate file.  Did that solve some particular problem?

mikerenfro:
The separate fix_homedir_*.yml files? I'd assumed that was preferred, but I'm not an expert. So if it's just as clean to roll them into the main playbook, I'm ok with that.

tasks/create_homedir.yml
```diff
@@ -0,0 +1,6 @@
+---
+
+- name: "Creating home directory for {{ item.split(':')[0] }}"
+  file:
+    path: "{{ item.split(':')[2] }}"
+    state: directory
```
jamescassell: I'm not sure whether we want to create homedirs like this; usually they would at least include /etc/skel.  At least with rhel6, you wanted oddjob-mkhomedir to do it for you because of selinux issues... anyone know if this is still an issue?

tasks/fix-cat2.yml
```diff
@@ -822,8 +822,8 @@
       - notimplemented
```
jamescassell: I'd clear out these 'notimplemented' tags

tasks/create_homedir.yml
mikerenfro: I won't claim this particular task is good for anything other than ticking off a checkbox that the home directory exists. Realistically, any normal system wouldn't have any of these to fix.

tasks/fix-cat2.yml
```diff
@@ -811,48 +811,55 @@
...
+    state: directory
+  with_items: "{{ rhel7stig_passwd_homedir_check.stdout_lines }}"
   when: rhel_07_020650
   tags:
       - cat2
       - medium
       - patch
-      - RHEL-07-020670
-      - notimplemented
+      - RHEL-07-020650
```
jamescassell: Good catch. strange how the tags are mis-named...

tasks/fix-cat2.yml
```diff
@@ -811,48 +811,55 @@
       - home
 
 - name: "MEDIUM | RHEL-07-020620 | PATCH | All local interactive user home directories defined in the /etc/passwd file must exist."
-  command: "true"
-  changed_when: no
+  file:
+    path: "{{ item.split(':')[2] }}"
+    state: directory
+  with_items: "{{ rhel7stig_passwd_homedir_check.stdout_lines }}"
```
jamescassell: I'd be okay with this if you add "check_mode: yes" and change PATCH to AUDIT in the name.

tasks/fix-cat2.yml
```diff
@@ -811,48 +811,55 @@
       - home
 
 - name: "MEDIUM | RHEL-07-020620 | PATCH | All local interactive user home directories defined in the /etc/passwd file must exist."
```
jamescassell: Change PATCH to AUDIT

tasks/fix-cat2.yml
```diff
@@ -811,48 +811,55 @@
...
-  changed_when: no
+  file:
+    path: "{{ item.split(':')[2] }}"
+    state: directory
+  with_items: "{{ rhel7stig_passwd_homedir_check.stdout_lines }}"
   when: rhel_07_020620
   tags:
       - cat2
       - medium
       - patch
```
jamescassell: swap out for audit

tasks/fix-cat2.yml
mikerenfro: I thought I saw an earlier note that tags would be deprecated in the long run, but no reason to have them wrong in the interim.

tasks/prelim.yml
```diff
@@ -135,6 +135,13 @@
       - audit
       - RHEL-07-020330
 
+- name: "PRELIM | RHEL-07-020620 RHEL-07-020630 RHEL-07-021640 RHEL-07-021650 | Check /etc/passwd entries for local user home directories"
+  shell: "egrep ':[0-9]{4}' /etc/passwd | cut -d: -f1,4,6"
+  changed_when: no
+  failed_when: no
+  register: rhel7stig_passwd_homedir_check
+  when: rhel_07_020620 or rhel_07_020630 or rhel_07_020640 or rhel_07_020650
```
jamescassell: I might consider splitting this out as:

```
  when:
      - rhel_07_020620 or 
        rhel_07_020630 or 
        rhel_07_020640 or 
        rhel_07_020650
```

Similarly, add the corresponding tags.

tasks/fix-cat2.yml
```diff
@@ -811,48 +811,55 @@
...
+    state: directory
+  with_items: "{{ rhel7stig_passwd_homedir_check.stdout_lines }}"
   when: rhel_07_020640
   tags:
       - cat2
       - medium
       - patch
-      - RHEL-07-020660
-      - notimplemented
+      - RHEL-07-020640
```
jamescassell: Good catch.

tasks/fix-cat2.yml
jamescassell: I must have missed that note.  I find the per-task tags very helpful, especially during development of this role...  the 'cat2', 'medium', 'patch', 'audit', etc. are less useful to me.

tasks/fix-cat2.yml
mikerenfro: But this one's the actual fix, not the audit, right?

tasks/fix-cat2.yml
mikerenfro: Shouldn't check_mode: yes go in the preliminary task, though? If that's intended as a 'do this even on a dry run' flag.

tasks/fix-cat2.yml
jamescassell: @shepdelacreme , do we want to label tasks as "PATCH" if all the task does is warn the user?

tasks/prelim.yml
```diff
@@ -139,8 +139,21 @@
   shell: "egrep ':[0-9]{4}' /etc/passwd | cut -d: -f1,4,6"
   changed_when: no
   failed_when: no
-  register: rhel7stig_passwd_homedir_check
-  when: rhel_07_020620 or rhel_07_020630 or rhel_07_020640 or rhel_07_020650
+  check_mode: yes
```
jamescassell: should be "check_mode: no" which causes the command to execute, even if check_mode is otherwise in effect.  "command" and "shell"  tasks are normally skipped in check mode because ansible can't know if it will change something.  That's why travis-ci failed.

tasks/fix-cat2.yml
mikerenfro: As written, these lines for 020620 through 020650 are file tasks, not just debug warnings. Though if warnings should be flagged as AUDIT or something else, I'll make sure to add that to the sudoers PR.

tasks/fix-cat2.yml
jamescassell: "check_mode: yes" means "make sure this task does not change anything, i.e., 'run it in check mode always'"  That task would always execute as if "--check" had been specified on the command line, even if the play is otherwise not running in check mode.

tasks/fix-cat2.yml
shepdelacreme: This is a good question. When we deprecated the separate AUDIT tasks we sort of did away with that nomenclature. However I don’t think we ever set a standard. For tasks that don’t make any changes I.e. debug warnings or tasks that perform checks prior to applying fix, it may be preferable to use AUDIT...but I’m open to other preferences.

shepdelacreme:
Thanks for all these changes and the other PRs as well! I have limited connectivity until late this week so I may not have a chance to do a full review/merge for these until Fri/Sat.

mikerenfro:
No problem. I'm just trying to get things into a semblance of working for the subset of rules we want for our servers. Won't likely bother with GUIs or smartcards, for example.

PR #73, opened 2018-03-20
Working on fixes for RHEL-07-010340 and RHEL-07-010350.

mikerenfro:


tasks/prelim.yml
```diff
@@ -92,6 +92,20 @@
       - audit
       - RHEL-07-010240
 
+- name: "PRELIM | Check /etc/sudoers and /etc/sudoers.d/* entries for NOPASSWD."
```
jamescassell: I wouldn't put these in prelim.yml, I'd put them as part of the task within a named "block:", since these aren't needed in multiple tasks, though I know it's not consistent throughout the codebase.

tasks/fix-cat2.yml
```diff
@@ -485,7 +485,9 @@
       - pamd
 
 - name: "MEDIUM | RHEL-07-010340 | PATCH | Users must provide a password for privilege escalation."
-  command: "true"
+  debug:
+    msg: "NOPASSWD entry for sudo: {{ item }}"
```
jamescassell: okay, but better would be a remediation that actually fixes the issue.

tasks/prelim.yml
```diff
@@ -92,6 +92,20 @@
...
+  shell: "grep '^[^#].*NOPASSWD' /etc/sudoers /etc/sudoers.d/*"
+  changed_when: no
+  failed_when: no
+  register: rhel_07_010340_audit
+  when: rhel_07_010340
+
+- name: "PRELIM | Check /etc/sudoers and /etc/sudoers.d/* entries for !authenticate."
+  shell: "grep '^[^#].*!authenticate' /etc/sudoers /etc/sudoers.d/*"
+  changed_when: no
+  failed_when: no
```
jamescassell: This should check for rc>1, as grep can fail.

tasks/prelim.yml
```diff
@@ -92,6 +92,20 @@
       - audit
       - RHEL-07-010240
 
+- name: "PRELIM | Check /etc/sudoers and /etc/sudoers.d/* entries for NOPASSWD."
+  shell: "grep '^[^#].*NOPASSWD' /etc/sudoers /etc/sudoers.d/*"
+  changed_when: no
+  failed_when: no
```
jamescassell: needs "check_mode: no"

tasks/fix-cat2.yml
```diff
@@ -485,7 +485,9 @@
       - pamd
 
 - name: "MEDIUM | RHEL-07-010340 | PATCH | Users must provide a password for privilege escalation."
```
jamescassell: if we're not actually fixing the issue, we probably just want to call it "AUDIT"

tasks/fix-cat2.yml
```diff
@@ -496,10 +498,11 @@
       - patch
       - RHEL-07-010340
       - sudoers
-      - notimplemented
 
 - name: "MEDIUM | RHEL-07-010350 | PATCH | Users must re-authenticate for privilege escalation."
```
jamescassell: perhaps something like:
```
- name: "MEDIUM | RHEL-07-010350 | Users must re-authenticate for privilege escalation."
  block:
      - name:  "MEDIUM | RHEL-07-010350 | AUDIT | Users must re-authenticate for privilege escalation."
        shell: "grep '^[^#].*!authenticate' /etc/sudoers /etc/sudoers.d/*"
        changed_when: no
        failed_when: rhel_07_010350_audit.rc > 1
        register: rhel_07_010350_audit

      - name: "MEDIUM | RHEL-07-010350 | AUDIT | Users must re-authenticate for privilege escalation."
        debug:
            msg: "!authenticate entry for sudo: {{ item }}"
```
[...]

tasks/fix-cat2.yml
mikerenfro: I'm good with a remediation, but I wasn't prepared to risk locking people out entirely by removing items. If replacing all instances of `NOPASSWD` with `PASSWD` is a guaranteed fix, I'm good with that, too.

tasks/fix-cat2.yml
```diff
@@ -485,10 +485,11 @@
...
 
 - name: "MEDIUM | RHEL-07-010340 | PATCH | Users must provide a password for privilege escalation."
-  debug:
-    msg: "NOPASSWD entry for sudo: {{ item }}"
-  with_items: "{{ rhel_07_010340_audit.stdout_lines }}"
-  changed_when: no
+  replace:
+    path: "{{ item }}"
+    regexp: '^([^#].*)NOPASSWD(.*)'
+    replace: '\1PASSWD\2'
```
jamescassell: Looks great!  Have you tested it?

tasks/fix-cat2.yml
```diff
@@ -500,10 +501,11 @@
...
 
 - name: "MEDIUM | RHEL-07-010350 | PATCH | Users must re-authenticate for privilege escalation."
-  debug:
-    msg: "!authenticate entry for sudo: {{ item }}"
-  with_items: "{{ rhel_07_010350_audit.stdout_lines }}"
-  changed_when: no
+      replace:
+        path: "{{ item }}"
+        regexp: '^([^#].*)!authenticate(.*)'
+        replace: '\1authenticate\2'
```
jamescassell: looks good

tasks/fix-cat2.yml
```diff
@@ -485,10 +485,11 @@
       - pamd
 
 - name: "MEDIUM | RHEL-07-010340 | PATCH | Users must provide a password for privilege escalation."
-  debug:
-    msg: "NOPASSWD entry for sudo: {{ item }}"
-  with_items: "{{ rhel_07_010340_audit.stdout_lines }}"
-  changed_when: no
+  replace:
+    path: "{{ item }}"
```
jamescassell: with sudoers type files, we definitely want 'validate: visudo -cf %s'

tasks/prelim.yml
```diff
@@ -92,6 +92,14 @@
       - audit
       - RHEL-07-010240
 
+- name: "PRELIM | Find all sudoers files."
+  command: "find /etc/sudoers /etc/sudoers.d/ -type f"
```
jamescassell: we should exclude files that end in '~' and files that contain '.' since sudo ignores those.

tasks/fix-cat2.yml
mikerenfro: Tested in practice? No. But PASSWD is the complementary flag to NOPASSWD, and it does the correct search and replace.

tasks/fix-cat2.yml
mikerenfro: For some reason, this one also grabs `!authenticate` in commented lines, even though what should be an equivalent `NOPASSWD` replace above correctly ignores comments.

shepdelacreme:
This looks good. Leaving this note for future reference:
**These checks will cause issues when applied to AWS/Cloud or other images that rely on private key only authentication with all user passwords locked/set to invalid.**

PR #74, opened 2018-03-20
Implementing 040720.

mikerenfro:
Implementing RHEL-07-040720.
Modifying RHEL-07-040700 to install tftp and tftp server when required.

### DIFF
--- a/tasks/fix-cat1.yml
+++ b/tasks/fix-cat1.yml
@@ -408,9 +408,8 @@
       name:
           - tftp
           - tftp-server
-      state: absent
+      state: "{{rhel7stig_tftp_required | ternary('present', 'absent') }}"
   when:
-      - not rhel7stig_tftp_required
       - rhel_07_040700
   tags:
       - cat1
--- a/tasks/fix-cat1.yml
+++ b/tasks/fix-cat1.yml
@@ -410,8 +410,8 @@
           - tftp-server
       state: absent
   when:
-      - rhel_07_040700
       - not rhel7stig_tftp_required
+      - rhel_07_040700
   tags:
       - cat1
       - high
--- a/tasks/fix-cat1.yml
+++ b/tasks/fix-cat1.yml
@@ -408,9 +408,10 @@
       name:
           - tftp
           - tftp-server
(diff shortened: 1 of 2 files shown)


tasks/fix-cat1.yml
```diff
@@ -408,9 +408,8 @@
       name:
           - tftp
           - tftp-server
-      state: absent
+      state: "{{rhel7stig_tftp_required | ternary('present', 'absent') }}"
```
jamescassell: I don't think we should install anything here.

tasks/fix-cat2.yml
```diff
@@ -1952,9 +1952,15 @@
...
 - name: "MEDIUM | RHEL-07-040720 | PATCH | If the Trivial File Transfer Protocol (TFTP) server is required, the TFTP daemon must be configured to operate in secure mode."
-  command: "true"
-  changed_when: no
-  when: rhel_07_040720
+  lineinfile:
+    path: /etc/xinetd.d/tftp
+    regexp: "(?i)^.*server_args.*="
+    line: "\tserver_args\t\t= -s /var/lib/tftpboot"
+    insertafter: "\tserver\t\t\t="
+    state: present
```
jamescassell: then I'd add:
```
register: result
failed_when:
    - result | failed
    - result.rc != 257
```
so you can just ignore the case where the file is not present

tasks/fix-cat1.yml
```diff
@@ -408,9 +408,8 @@
...
   when:
-      - not rhel7stig_tftp_required
```
jamescassell: I'd leave this.

tasks/fix-cat1.yml
shepdelacreme: I concur I don’t think the role should install these packages when the “required” variable is set to true. Should just remove when required = false. The package/service install is up to another role.

tasks/fix-cat1.yml
```diff
@@ -410,8 +410,8 @@
           - tftp-server
       state: absent
   when:
-      - not rhel7stig_tftp_required
       - rhel_07_040700
+      - not rhel7stig_tftp_required
```
jamescassell: I'd prefer not to relocate this line...

PR #76, opened 2018-03-20
Working on RHEL-07-040680.

mikerenfro:
Checking client restrictions with `postconf -n`, adjusting with `postconf -e` if anything other than `permit_mynetworks, reject`.

### DIFF
--- a/tasks/fix-cat2.yml
+++ b/tasks/fix-cat2.yml
@@ -1941,15 +1941,15 @@
       - RHEL-07-040670
 
 - name: "MEDIUM | RHEL-07-040680 | PATCH | The system must be configured to prevent unrestricted mail relaying."
-  command: "true"
-  changed_when: no
-  when: rhel_07_040680
+  command: "postconf -e 'smtpd_client_restrictions=permit_mynetworks, reject'"
+  when:
+    - rhel_07_040680
+    - rhel_07_040680_audit.stdout != 'smtpd_client_restrictions = permit_mynetworks, reject'
   tags:
       - cat2
       - medium
       - patch
       - RHEL-07-040680
-      - notimplemented
 
 - name: "MEDIUM | RHEL-07-040720 | PATCH | If the Trivial File Transfer Protocol (TFTP) server is required, the TFTP daemon must be configured to operate in secure mode."
   command: "true"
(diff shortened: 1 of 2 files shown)


tasks/fix-cat2.yml
```diff
@@ -1941,15 +1941,15 @@
       - RHEL-07-040670
 
 - name: "MEDIUM | RHEL-07-040680 | PATCH | The system must be configured to prevent unrestricted mail relaying."
-  command: "true"
-  changed_when: no
-  when: rhel_07_040680
+  command: "postconf -e 'smtpd_client_restrictions=permit_mynetworks, reject'"
```
jamescassell: Can we assume postfix is installed?  It is by default, but it's a leaf package, so can be easily removed.

mikerenfro: Probably shouldn't assume it, no. Since the STIG says to check if it's installed, and if it's not, 040680 is not applicable.

tasks/fix-cat2.yml
```diff
@@ -1940,11 +1940,22 @@
...
 
-- name: "MEDIUM | RHEL-07-040680 | PATCH | The system must be configured to prevent unrestricted mail relaying."
-  command: "postconf -e 'smtpd_client_restrictions=permit_mynetworks, reject'"
-  when:
-    - rhel_07_040680
-    - rhel_07_040680_audit.stdout != 'smtpd_client_restrictions = permit_mynetworks, reject'
+- block:
+    - name: "MEDIUM | RHEL-07-040680 | AUDIT | Check if postfix is installed."
+      command: rpm -q postfix
+      failed_when: no
```
jamescassell: You'll also need "check_mode: no" and "changed_when: no"

tasks/fix-cat2.yml
```diff
@@ -1940,11 +1940,22 @@
...
+      register: rhel_07_040680_rpm_audit
+
+    - name: "MEDIUM | RHEL-07-040680 | AUDIT | The system must be configured to prevent unrestricted mail relaying."
+      command: "postconf -n smtpd_client_restrictions"
+      changed_when: no
```
jamescassell: plus "check_mode: no"

PR #78, opened 2018-03-21
Implementing RHEL-07-031010

mikerenfro:


defaults/main.yml
```diff
@@ -304,6 +304,9 @@ rhel7stig_time_service_configs:
 
 rhel7stig_firewall_service: firewalld
 
+# RHEL-07-031010
+rhel7stig_log_aggregator: false
```
jamescassell: how about rhel7stig_system_is_log_aggregator, similar to rhel7stig_system_is_router above?

tasks/fix-cat2.yml
```diff
@@ -1304,16 +1304,27 @@
       - notimplemented
 
 - name: "MEDIUM | RHEL-07-031010 | PATCH | The rsyslog daemon must not accept log messages from other servers unless the server is being used for log aggregation."
-  command: "true"
-  changed_when: no
-  when: rhel_07_031010
+  replace:
+    path: /etc/rsyslog/rsyslog.conf
```
jamescassell: should be /etc/rsyslog.conf

tasks/fix-cat2.yml
```diff
@@ -1304,16 +1304,27 @@
...
+    regexp: '({{ item }})'
+    replace: '#\1'
+    register: result
+  with_items:
+    - '^(\$ModLoad imtcp)'
```
jamescassell: I see that the STIG checks for tmtcp.  Where did you get the others below?

tasks/fix-cat2.yml
```diff
@@ -1304,16 +1304,27 @@
...
+    - '^(\$ModLoad imudp)'
+    - '^(\$UDPServerRun)'
+    - '^(\$InputTCPServerRun)'
+  when:
+    - rhel_07_031010
+    - not rhel7stig_log_aggregator
+  failed_when:
```
jamescassell: I think you can get rid of the failed_when because rsyslog should always be installed, which provides /etc/rsyslog.conf (and if not, we can add a variable saying rsyslog not required.)  otherwise, at least add "result | failed" to the conditions here to catch the case where a successful replacement is done, or no replacement needs to be done.

defaults/main.yml
mikerenfro: Sure.

PR #84, opened 2018-03-22
Working on 030320 and 030321.

mikerenfro:


tasks/fix-cat2.yml
```diff
@@ -1144,26 +1144,48 @@
...
+        name: audispd-plugins
+        state: present
+
+    - name: "Set disk_full_action"
+      lineinfile:
+        path: /etc/audisp/audisp-remote.conf
+        regexp: "disk_full_action =.*"
+        line: "disk_full_action = {{ rhel7stig_audisp_disk_full_action }}"
+    # Technically, network_failure_action should only be in 030321, but it's also called out in 030320.
+    - name: "Set network_failure_action"
```
shepdelacreme: I think we should remove the network_failure_action task from this block. Technically it is in the 030320 STIG ID fix text but it seems like an error that should be fixed in a later STIG revision. Its just a duplicative task in this role now.

PR #85, opened 2018-03-22
Working on 030330.

mikerenfro:


tasks/fix-cat2.yml
```diff
@@ -1166,15 +1166,20 @@
...
-  command: "true"
-  changed_when: no
+  lineinfile:
+    path: /etc/audit/auditd.conf
+    regexp: "space_left =.*"
+    line: "space_left = {{ rhel7stig_auditd_space_left }}"
+  register: rhel7stig_auditd_space_left_result
+  failed_when:
+    - rhel7stig_auditd_space_left_result | failed
+    - rhel7stig_auditd_space_left_result.rc != 257
```
jamescassell: I think we only want to ignore the case of the file being absent when an absent file satisfies the STIG requirement.

jamescassell:
See #87

PR #87, opened 2018-03-22
Audit config

jamescassell:
slight overlap with existing PRs

Please don't squash.

### DIFF
--- a/tasks/fix-cat2.yml
+++ b/tasks/fix-cat2.yml
@@ -1111,15 +1111,17 @@
       - notimplemented
 
 - name: "MEDIUM | RHEL-07-030010 | PATCH | The operating system must shut down upon audit processing failure, unless availability is an overriding concern. If availability is a concern, the system must alert the designated staff (System Administrator [SA] and Information System Security Officer [ISSO] at a minimum) in the event of an audit processing failure."
-  command: "true"
-  changed_when: no
+  lineinfile:
+      path: /etc/audit/rules.d/audit.rules
+      regexp: "^-f "
+      line: "-f {{ rhel7stig_auditd_failure_flag }}"
+  notify: update running audit failure mode
   when: rhel_07_030010
   tags:
       - cat2
       - medium
       - patch
       - RHEL-07-030010
-      - notimplemented
 
 - name: "MEDIUM | RHEL-07-030300 | PATCH | The operating system must off-load audit records onto a different system or media from the system being audited."
   command: "true"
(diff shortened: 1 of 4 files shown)


tasks/fix-cat2.yml
```diff
@@ -1166,15 +1168,18 @@
       - notimplemented
 
 - name: "MEDIUM | RHEL-07-030330 | PATCH | The operating system must immediately notify the System Administrator (SA) and Information System Security Officer ISSO (at a minimum) when allocated audit record storage volume reaches 75% of the repository maximum audit record storage capacity."
-  command: "true"
-  changed_when: no
+  lineinfile:
+      path: /etc/audit/auditd.conf
+      regexp: ^space_left\s*=\s*
+      line: space_left = {{ ( ansible_mounts | json_query(query) / 1024 / 1024 ) | int + 1 }}
```
shepdelacreme: What is the outcome of this? It looks like it converts the volume size to MB then converts that to an `integer` type and adds 1? So if the volume is 500MB it would set the space left to 501MB? or am I not reading this right. 

If we take this route of automatically gathering the /var/log/audit volume size we should do something like `{{ ( ansible_mounts | json_query(query) * 0.25 / 1024 / 1024 ) | int }}` to get 25% of the total size?

Note: not sure if the above filter syntax actually works its just an example.

I also like the method @mikerenfro uses in #85 allowing a user to set a static value in vars file. Maybe we can move this entire filter line to the vars file making it easier for a user to override if they want? i.e.

in `defaults/main.yml` set up a var like in #85. 
```
# RHEL-07-030330: Use the automatic space left calculation based on /var/log/audit volume size or set this to a custom value that is at least 25% of the free space in /var/log/audit (measured in megabytes)
rhel7stig_auditd_space_left: {{ ( ansible_mounts | json_query(query) * 0.25 / 1024 / 1024 ) | int }}
```

Then in this task:
```
line: space_left = {{ rhel7stig_auditd_space_left }}

jamescassell: Good catch, it does look like I didn't use 25%, oops. I'm fine moving it to the defaults file.

jamescassell: If you want to merge #85, I can rebase this PR.

shepdelacreme: Ok I merged #85

jamescassell:
rebase in progress

jamescassell:
Please see #91

PR #88, opened 2018-03-22
STIG ID 021620 - AIDE rule - sha512 setting

mikerenfro:


tasks/fix-cat2.yml
```diff
@@ -1089,15 +1089,31 @@
...
+    - name: "Replace sha256+sha512 entries with sha512"
+      replace:
+        path: /etc/aide.conf
+        regexp: '([A-Z]+ = .*)(sha256\+sha512)(.*)'
+        replace: '\1sha512\3'
+      register: rhel_07_021620_sha256_512_result
+      failed_when:
+        - rhel_07_021620_sha256_512_result | failed
+        - rhel_07_021620_sha256_512_result.rc != 257
+    - name: "Replace sha256 entries with sha512"
```
shepdelacreme: Can you add a space above this line for readability.

tasks/fix-cat2.yml
```diff
@@ -1089,15 +1089,31 @@
...
+      replace:
+        path: aide.conf
```
shepdelacreme: This should be `/etc/aide.conf` I think?

mikerenfro: Yep. Bad test environment. Thought I had all those fixed. Will adjust.

PR #89, opened 2018-03-23
Fix OS-level FIPS crypto

jamescassell:
These changes remediate a system that was not built with fips=1 on the installer command line.

Please don't squash.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -483,3 +483,4 @@ rhel7stig_audit_part: "{{ rhel_07_audit_part.stdout }}"
 rhel7stig_boot_part: "{{ rhel_07_boot_part.stdout }}"
 
 rhel7stig_machine_uses_uefi: "{{ rhel_07_sys_firmware_efi.stat.exists }}"
+rhel7stig_grub_cfg_path: "{{ rhel7stig_machine_uses_uefi | ternary('/boot/efi/EFI/{{ ansible_distribution | lower }}/grub.cfg', '/boot/grub2/grub.cfg') }}"
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -481,3 +481,5 @@ rhel7stig_auditd_failure_flag: "{{ rhel7stig_availability_override | ternary(1, 
 rhel7stig_audit_part: "{{ rhel_07_audit_part.stdout }}"
 
 rhel7stig_boot_part: "{{ rhel_07_boot_part.stdout }}"
+
+rhel7stig_machine_uses_uefi: "{{ rhel_07_sys_firmware_efi.stat.exists }}"
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -479,3 +479,5 @@ rhel7stig_availability_override: false
 rhel7stig_auditd_failure_flag: "{{ rhel7stig_availability_override | ternary(1, 2) }}"
 
 rhel7stig_audit_part: "{{ rhel_07_audit_part.stdout }}"
+
+rhel7stig_boot_part: "{{ rhel_07_boot_part.stdout }}"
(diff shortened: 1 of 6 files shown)


jamescassell:
I've wasted too much time trying to workaround the travis-ci failures.  Suggestions welcome.

jamescassell:
ended up using the rhel7stig_skip_for_travis variable and skipping the rebuild grub task, even though I think it's ugly.

shepdelacreme:
There is some serious dark magic in this PR with the regex parameter stuff. 

Since this is a rather large set of changes, holding off till most of the other stuff touching some of these files is merged. Once all the smaller changes are sorted I'll request a rebase on this one.

jamescassell:
Sounds fine.  Let me know when you're ready for a rebase.

shepdelacreme:
@jamescassell can you rebase this PR

Thanks!

jamescassell:
Since there were no comments on the original commits, I'm rebasing and force-pushing to this same PR.  The original commits from this PR are available here: https://github.com/jamescassell/RHEL7-STIG/commits/fips-crypto-pr89v1

shepdelacreme:
@jamescassell I'm reviewing the STIG V1R4 that was released on Jan 26th 2018 and it looks like they removed the bit about `prelink` being disabled from the Fix text on 021350. Do you want to remove those prelink tasks from this PR or let that get removed in a later PR?

jamescassell:
I looked and see what you are saying.  Here's Red Hat's instructions for enabling fips mode: https://access.redhat.com/solutions/137833

They specifically say to disable prelinking:
"4. If the prelink package is installed: disable prelinking (uninstalling the prelink package is optional)"

I haven't tested if fips mode works with prelinking enabled.  I wouldn't want to leave someone with a broken system by attempting to enable fips mode if that inherently requires disabled prelinking.  Needs testing, in my opinion.  My kickstarts don't install prelink.

On a related note, the 'undo prelinking' task as written is not idempotent, but I think it'd be hard to make it so.  Perhaps by moving it to handlers and adding a notify to the /etc/sysconfig/prelink task... but if these tasks aren't needed, we should remove them.

jamescassell:
The STIG V1R4 Revision History says:
"- Updated RHEL-07-021350: Removed wording from the Fix that indicated "prelink" must be disabled and the instructions about GRUB configuration."

shepdelacreme:
Yeah I did a diff on the two versions and the only thing changed was the removal of the following line from the fix text:

> For proper operation of the in-module integrity verification, the prelink has to be disabled. This can be done by configuring PRELINKING=no in the "/etc/sysconfig/prelink" configuration file. Existing prelinking, if any, should be undone on all system files using the prelink -u -a command.

I see what you are saying in the RedHat instructions. I'm curious why DISA removed that requirement from the fix. It seems that with RHEL7 the default is to not install prelink. See step 4 on this doc https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/security_guide/chap-federal_standards_and_regulations

> Make sure prelinking is disabled.
For proper operation of the in-module integrity verification, prelinking of libraries and binaries has to be disabled. Prelinking is done by the prelink package, which is not installed by default. Unless prelink has been installed, this step is not needed. To disable prelinking, set the PRELINKING=no option in the /etc/sysconfig/prelink configuration file. To disable existing prelinking on all system files, use the prelink -u -a command.

Maybe to solve the non-idempotent problem we simply make the prelink package check the first task. If that detects that prelink is installed we do 1 of two things:

1. Undo prelinking and uninstall prelink completely OR
1. Disable prelink by editing /etc/sysconfig/prelink and then undo existing prelinking

Making it conditional on if the prelink package is installed or making it a handler makes sense. I also think we shouldn't create `/etc/sysconfig/prelink` if the package isn't installed?

jamescassell:
Pushed another commit that implements your option 2.

shepdelacreme:
Merged thanks!

PR #90, opened 2018-03-23
STIG IDs 021600 and 021610 - AIDE rules - xattrs and acl settings

mikerenfro:
Using regexp instead of replace for LOG comment and LOG entry, even though there's only one of each. lineinfile would end up tacking extra lines at the end of the file. Maybe there's a clever way to add xattrs if and only if they're not already on the line.


tasks/fix-cat3.yml
```diff
@@ -89,27 +89,54 @@
       - mount
       - tmp
 
-- name: "LOW | RHEL-07-021600 | PATCH | The file integrity tool must be configured to verify Access Control Lists (ACLs)."
-  command: "true"
-  changed_when: no
+- block:
+    - name: "LOW | RHEL-07-021600 | PATCH | The file integrity tool must be configured to verify Access Control Lists (ACLs)."
```
jamescassell: We should clarify the code guidelines for indentation.  I see that @shepdelacreme has used two additional spaces before the '-' inside a block, and also 4 additional spaces, all in different parts of the same commit.  my reading of the code guidelines (which says 4 spaces) suggests 4 additional spaces, which is what we do for "tags", "when", etc.

mikerenfro: Will see what it takes to get emacs to make this more automatic. Or will just do it the manual way.

mikerenfro: [yamllint](https://github.com/adrienverge/yamllint), perhaps? No idea if it's comprehensive enough for everything, but it may handle many of the style fixes.

shepdelacreme: Sorry for the confusion with indentation here. The original style guide stated 4 spaces. However nothing has automated the checking/enforcement of that so its not consistent everywhere. For now strive for 4 spaces because its in the style guide and I think it is more readable. I will work on getting `yamllint` or `ansible-lint` setup on this repo and see if we can enforce it as part of the CI checks.

tasks/fix-cat3.yml
```diff
@@ -90,26 +90,76 @@
...
+    - name: "Replace CONTENT with CONTENT_EX"
+      replace:
+        path: /etc/aide.conf
+        regexp: '^(.*) (CONTENT)$'
+        replace: '\1 CONTENT_EX'
+      register: result
+      failed_when:
+        - result | failed
+        - result.rc != 257
+    - name: "Replace NORMAL with DATAONLY"
```
shepdelacreme: Please add a space above this line for readability.

tasks/fix-cat3.yml
```diff
@@ -90,26 +90,76 @@
...
+      replace:
+        path: /etc/aide.conf
+        regexp: '^(.*) (NORMAL)$'
+        replace: '\1 DATAONLY'
+      register: result
+      failed_when:
```
shepdelacreme: It might be worth it to have a `stat` task in this block and the same in the block in #88 to check if `/etc/aide.conf` exists and then only run these tasks if it exists. That would avoid the custom `failed_when` here.

shepdelacreme:
Reading the STIG for 021600 and 021610 I'm not sure these replace tasks are meeting the remediation requirements, but maybe I'm just not reading it right. 

**Here is the relevant fix text I have from 021600:**
**Check Text**: Verify the file integrity tool is configured to verify ACLs.

Check the "aide.conf" file to determine if the "acl" rule has been added to the rule list being applied to the files and directories selection lists.

An example rule that includes the "acl" rule is below:

All= p+i+n+u+g+s+m+S+sha512+acl+xattrs+selinux
/bin All # apply the custom rule to the files in bin 
/sbin All # apply the same custom rule to the files in sbin 

If the "acl" rule is not being used on all selection lines in the "/etc/aide.conf" file, or ACLs are not being checked by another file integrity tool, this is a finding.

**Fix Text**: Configure the file integrity tool to check file and directory ACLs. 

If AIDE is installed, ensure the "acl" rule is present on all file and directory selection lists.

**And for 021610:**
**Check Text**: Verify the file integrity tool is configured to verify extended attributes.

Check the "aide.conf" file to determine if the "xattrs" rule has been added to the rule list being applied to the files and directories selection lists.

An example rule that includes the "xattrs" rule follows:

All= p+i+n+u+g+s+m+S+sha512+acl+xattrs+selinux
/bin All # apply the custom rule to the files in bin 
/sbin All # apply the same custom rule to the files in sbin 

If the "xattrs" rule is not being used on all selection lines in the "/etc/aide.conf" file, or extended attributes are not being checked by another file integrity tool, this is a finding.

**Fix Text**: Configure the file integrity tool to check file and directory extended attributes. 

If AIDE is installed, ensure the "xattrs" rule is present on all file and directory selection lists.

mikerenfro:
The logic was to remediate an out of the box install of aide with as few changes as possible, and that anyone who had their own aide rules could just disable these tasks.

The following default rules included both `acl` and `xattrs`: `DIR`, `PERMS`, `STATIC`, `CONTENT_EX`, `DATAONLY`. So any selection iist using any of those rules will comply.

The default inclusion lists use `CONTENT`, `CONTENT_EX`, `DATAONLY`, `LOG`, `NORMAL`, and `PERMS` rules.

As `PERMS`, `CONTENT_EX`, and `DATAONLY` are already on the list of complying rules, we have the option to change any `CONTENT`, `LOG`, and `NORMAL` inclusions to the closest complying equivalent. That seemed to be `CONTENT` -> `CONTENT_EX`, and `NORMAL` -> `DATAONLY`.

As for `LOG`, I just added the `xattrs` flag to its default rule, as I didn't think any of the others were that close. But I'm sure there are other ways to accomplish this.

shepdelacreme:
Closing outdated.

PR #96, opened 2018-03-26
Add/Remove STIG IDs based on STIG release 4

shepdelacreme:
- This removes STIG ID **RHEL-07-041004** and **RHEL-07-010080** per Version 1 Release 4.
- This adds placeholder tasks and vars for STIG ID **RHEL-07-010062** and **RHEL-07-010101** per Version 1 Release 4.


jamescassell:
looks good to me

PR #99, opened 2018-03-26
housekeeping and cleanup

jamescassell:
Various cleanup items.  The commit messages have more details.


tasks/main.yml
```diff
@@ -12,35 +12,31 @@
...
+- import_tasks: prelim.yml
   become: yes
   tags:
       - prelim_tasks
-      - always
 
 - name: Include CAT I patches
-  include: fix-cat1.yml
+  import_tasks: fix-cat1.yml
   static: no
```
shepdelacreme: Can you remove the `static: no` line as well. It is no longer needed.

tasks/main.yml
```diff
@@ -12,35 +12,31 @@
...
   when: rhel7stig_cat1_patch | bool
   tags:
       - cat1
-      - patch
       - high
 
 - name: Include CAT II patches
-  include: fix-cat2.yml
+  import_tasks: fix-cat2.yml
   static: no
```
shepdelacreme: Can you remove the `static: no` line as well. It is no longer needed.

tasks/main.yml
```diff
@@ -12,35 +12,31 @@
...
   when: rhel7stig_cat2_patch | bool
   tags:
       - cat2
-      - patch
       - medium
 
 - name: Include CAT III patches
-  include: fix-cat3.yml
+  import_tasks: fix-cat3.yml
   static: no
```
shepdelacreme: Can you remove the `static: no` line as well. It is no longer needed.

PR #101, opened 2018-03-26
Updating RHEL-07-010480 and RHEL-07-010490 to V1R4

mikerenfro:
V1R4 of RHEL-07-010480 and RHEL-07-010490 require the grub superuser to be root.


jamescassell:
Interestingly, the scap-security-guide specifically checks that the superuser is NOT root.

shepdelacreme:
Do we need this change? i.e. Do we want to remove the ability for the user to customize the boot superuser?

jamescassell:
I'd leave it as-is...

mikerenfro:
I'm not really invested in if it's set to root or not. Ours probably will be, just so I've got one less thing to explain. Closing a PR and deleting a branch is easy enough, so I'll defer to you two on that decision.

shepdelacreme:
Yeah I'm for closing this one out and leaving it as is. The role meets the intent of the STIG currently and also allows it to be customized for people that want that.

PR #103, opened 2018-03-26
Updating RHEL-07-021350 to V1R4 (prelinking is ok again)

mikerenfro:


jamescassell:
NAK.  See discussion in #89

mikerenfro:
Fair enough.

PR #104, opened 2018-03-26
Implementing RHEL-07-030360

mikerenfro:
Suggestions welcome for not having a separate audit_suid_sgid.yml, but I couldn't work out a way of getting audit_command's other parameters to show up correctly in an outer/inner loop structure where things like the STIG-ID and audit keyword are constant for all paths.


jamescassell:
check out with_nested: http://docs.ansible.com/ansible/2.4/playbooks_loops.html#nested-loops

mikerenfro:
I don't doubt that, but I'm not going to be up to one of those today. Also trying to figure out why the build failed.

jamescassell:
perhaps something like:
```
with_items: "{{ results | json_query('results[*].stdout_lines[]' }}"
```

mikerenfro:
@shepdelacreme Pretty sure this task is working correctly on my test box, so if you have a chance to dig through the Travis output to see why this is failing, I'd appreciate it.

(Granted, I now get an out of memory error during the audit_file portion, but I think that's unrelated.)

shepdelacreme:
@mikerenfro Ok I'll take a look and run some tests locally as well.

shepdelacreme:
I see you got this worked out! Looks good.

PR #106, opened 2018-03-27
Implementing RHEL-07-021030

mikerenfro:


shepdelacreme:
How safe is it to automatically change the group owner on directories that are found? I'm asking because I'm not sure myself but I feel like there is potential here for unintended outcomes.

mikerenfro:
That's the fix as given in the STIG, and maybe we file that under 'tool can break your system'.

Starting from what I assume is the common case of:

    drwxdrwxrwx 2 root badgroup 6 Mar 27 09:27 badfolder

if we remediate that to:

    drwxdrwxrwx 2 root root 6 Mar 27 09:27 badfolder

there shouldn't be any harm, as membership in `badgroup` didn't add any extra permissions. But if we had changed a:

    drwxdrwx-wx 2 root badgroup 6 Mar 27 09:27 badfolder

to:

    drwxdrwx-wx 2 root root 6 Mar 27 09:27 badfolder

then members of `badgroup` will lose read access to the contents. I can't imagine that's a likely scenario, but this can wait a while for further review.

jamescassell:
This one seems hard to remediate automatically.  What if a user chmod 777 a subdir in his homedir?  Yes, that's a STIG violation, but a better solution would be to chmod it to 775 or 755, in my opinion...

mikerenfro:
Still would be potential breakage if someone was relying on those 777 permissions, though.

Maybe this is a case of "[complying systems are all alike, non-complying systems are non-complying (and will be broken) in their own way](https://en.wikipedia.org/wiki/Anna_Karenina_principle)".

shepdelacreme:
Maybe we can have this PATCH operate as a warning when the STIG ID var is set to false? ie Maybe add a var for `rhel_07_021030_warn_only`. Then add a task in the block that prints a warning and have it and the remediation task execute when `rhel_07_021030_warn_only` is set to `true/false` respectively?

Or does that seem like overkill?

jamescassell:
How many remediations to we have that we /could/ fix, but might sometimes just want to warn?  If it's more than just one or two, would it be good to have a general variable to flip them all at once?

perhaps like:
```
rhel7stig_abnormal_warn_only: true
rhel_07_021030_warn_only: "{{ rhel7stig_abnormal_warn_only }}"
```

not sure what the best name for it would be, though...

shepdelacreme:
Once #133 gets rebased and merged we should update this task to use `rhel7stig_local_mounts` and then I'll merge this one.

mikerenfro:
Sure. Just shoot me a reminder when #133 is done.

shepdelacreme:
#133 has been merged. Go ahead and update/rebase. Thanks!

tasks/fix-cat2.yml
```diff
@@ -940,12 +940,21 @@
       - notimplemented
 
 - name: "MEDIUM | RHEL-07-021030 | PATCH | All world-writable directories must be group-owned by root, sys, bin, or an application group."
-  command: "true"
-  changed_when: no
+  block:
+    - name: "MEDIUM | RHEL-07-021030 | AUDIT | All world-writable directories must be group-owned by root, sys, bin, or an application group."
+      command: find {{ rhel7stig_local_mounts | join(' ') }} -type d -perm -002 -gid +999
```
jamescassell: Will want `-xdev`

tasks/fix-cat2.yml
```diff
@@ -940,12 +940,21 @@
...
+      changed_when: no
```
jamescassell: Change to: `rhel_07_021030_audit.stdout != ""`

defionscode:
@mikerenfro friendly ping, do you have cycles to make the updates requested by @jamescassell ?

tasks/fix-cat2.yml
mikerenfro: My apologies for the delay. Will get back into this and resolve it today if at all possible.

tasks/fix-cat2.yml
```diff
@@ -942,9 +942,9 @@
 - name: "MEDIUM | RHEL-07-021030 | PATCH | All world-writable directories must be group-owned by root, sys, bin, or an application group."
   block:
     - name: "MEDIUM | RHEL-07-021030 | AUDIT | All world-writable directories must be group-owned by root, sys, bin, or an application group."
-      command: find {{ rhel7stig_local_mounts | join(' ') }} -type d -perm -002 -gid +999
+      command: find {{ rhel7stig_local_mounts | join(' ') }} -xdev -type d -perm -002 -gid +999
```
jamescassell: Thanks!

tasks/fix-cat2.yml
```diff
@@ -942,9 +942,9 @@
...
       changed_when: no
-      register: rhel_07_021030_audit
+      register: rhel_07_021030_audit.stdout != ""
```
jamescassell: that should be on the "changed_when" line above, but I suppose the find command would also have to exclude items group-owned by root

mikerenfro:
Probably will have to do another rebase to catch back up with the current state of development.

tasks/fix-cat2.yml
mikerenfro: The find command's `-gid +999` should exclude any files owned by system-level users, including root, right?

jamescassell: You are probably correct.  I didn't test it.  In that case, it is correct to make:
```
changed_when: rhel_07_021030_audit.stdout != ""
```

and restore this line to:
```
register: rhel_07_021030_audit
```

jamescassell:
Should we give this the disruption-high treatment per concerns expressed above by @shepdelacreme ?

tasks/fix-cat2.yml
```diff
@@ -940,12 +940,21 @@
       - notimplemented
 
 - name: "MEDIUM | RHEL-07-021030 | PATCH | All world-writable directories must be group-owned by root, sys, bin, or an application group."
-  command: "true"
-  changed_when: no
+  block:
+    - name: "MEDIUM | RHEL-07-021030 | AUDIT | All world-writable directories must be group-owned by root, sys, bin, or an application group."
+      command: find {{ rhel7stig_local_mounts | join(' ') }} -xdev -type d -perm -002 -gid +999
+      changed_when: no
```
jamescassell: Also add:
```
check_mode: no
```
to the AUDIT task so check_mode still works.

jamescassell:
The failure is not your fault:
```
Installing:
 ansible                            noarch 2.6.0-1.el7            epel     10 M
```

The initial release of ansible-2.6.0 had a broken pamd module implementation.  ansible-2.6.1 fixes the issue.  2.6.1 is available in the Red Hat repos, or directly from ansible, but apparently not yet in epel.

mikerenfro:
I think this might have everything synced back up and everything fixed. My apologies for going dark for this long: many non-STIG-related tasks out this way since April.

tasks/fix-cat2.yml
```diff
@@ -1098,8 +1098,9 @@
   block:
       - name: "MEDIUM | RHEL-07-021030 | AUDIT | All world-writable directories must be group-owned by root, sys, bin, or an application group."
         command: find {{ rhel7stig_local_mounts | join(' ') }} -xdev -type d -perm -002 -gid +999
-        changed_when: no
-        register: rhel_07_021030_audit.stdout != ""
+        changed_when: rhel_07_021030_audit.stdout != ""
+        check_mode: "{{rhel7stig_disruptive_check_mode}}"
```
jamescassell: this should be `check_mode: no`

`{{ rhel7stig_disruptive_check_mode }}` should go on the following task that would actually make the change

tasks/fix-cat2.yml
```diff
@@ -1098,8 +1098,9 @@
...
+        register: rhel_07_021030_audit
 
       - name: "MEDIUM | RHEL-07-021030 | PATCH | All world-writable directories must be group-owned by root, sys, bin, or an application group."
```
jamescassell: add
```
check_mode: "{{ rhel7stig_disruptive_check_mode }}"
```

tasks/fix-cat2.yml
```diff
@@ -1110,6 +1111,7 @@
   when: rhel_07_021030
```
jamescassell: add
```
when:
    - rhel7stig_disruptive
```

tasks/fix-cat2.yml
mikerenfro: Ok. I must have misread the variable description and the other (single?) example to go from. Will fix these in the morning.

shepdelacreme:
Hmm CI is failing but I don't think it is related to these changes. Looks like it is loading Ansible 2.6 and there is an issue in the pamd_module. Putting this here for now so I can remember to investigate further.

```
TASK [role_under_test : MEDIUM | RHEL-07-010200 | PATCH | The PAM system service must be configured to store only encrypted representations of passwords.] ***
task path: /etc/ansible/roles/role_under_test/tasks/fix-cat2.yml:236
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: TypeError: 'bool' object is not callable
failed: [localhost] (item={u'state': u'args_present', u'args': [u'sha512']}) => {"changed": false, "item": {"args": ["sha512"], "state": "args_present"}, "module_stderr": "Traceback (most recent call last):\n  File \"/tmp/ansible__kny2g/ansible_module_pamd.py\", line 785, in <module>\n    main()\n  File \"/tmp/ansible__kny2g/ansible_module_pamd.py\", line 752, in main\n    valid, msg = service.validate()\n  File \"/tmp/ansible__kny2g/ansible_module_pamd.py\", line 656, in validate\n    if not current_line.is_valid()[0]:\nTypeError: 'bool' object is not callable\n", "module_stdout": "", "msg": "MODULE FAILURE", "rc": 1}
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: TypeError: 'bool' object is not callable
failed: [localhost] (item={u'state': u'args_absent', u'args': [u'md5', u'bigcrypt', u'sha256', u'blowfish']}) => {"changed": false, "item": {"args": ["md5", "bigcrypt", "sha256", "blowfish"], "state": "args_absent"}, "module_stderr": "Traceback (most recent call last):\n  File \"/tmp/ansible_hjb0tD/ansible_module_pamd.py\", line 785, in <module>\n    main()\n  File \"/tmp/ansible_hjb0tD/ansible_module_pamd.py\", line 752, in main\n    valid, msg = service.validate()\n  File \"/tmp/ansible_hjb0tD/ansible_module_pamd.py\", line 656, in validate\n    if not current_line.is_valid()[0]:\nTypeError: 'bool' object is not callable\n", "module_stdout": "", "msg": "MODULE FAILURE", "rc": 1}
```

jamescassell:
@shepdelacreme Yes, EPEL has a bad ansible version.  Here's the original bug report: https://github.com/ansible/ansible/issues/41179 And here's the patch that fixed it: https://github.com/ansible/ansible/pull/42025 A fixed build is here: https://bodhi.fedoraproject.org/updates/FEDORA-EPEL-2018-be727516a8

If you want to speed along the update on its journey to EPEL, you can add karma to that update.  Otherwise, it will automatically land in 11 days or so.

This PR needs rebasing, but otherwise looks fine to me.

mikerenfro:
@shepdelacreme Rebased and ready for pull, I think.

shepdelacreme:
Thanks reviewing!

shepdelacreme:
LGTM...hopefully the Ansible 2.6.1 bugfix release lands soon so tests pass again.

PR #109, opened 2018-03-27
remove time limit from aide init

jamescassell:
as written, aide init was killed after 45 seconds, which is
not enough time for it to run in our environments

turn it into a true fire-and-forget task

### DIFF
--- a/handlers/main.yml
+++ b/handlers/main.yml
@@ -24,11 +24,7 @@
       state: restarted
 
 - name: init aide
-  command: /usr/sbin/aide --init -B 'database_out=file:/var/lib/aide/aide.db.gz'
-  async: 45
-  poll: 0
-  tags:
-    - aide
+  shell: nohup /usr/sbin/aide --init -B 'database_out=file:/var/lib/aide/aide.db.gz' > /dev/null &
 
 - name: restart auditd
   command: service auditd restart


shepdelacreme:
Hmm I think we need a way to allow this task to operate in either mode. i.e. In our primary use of these roles we are building "golden images" and so we need to the AIDE init task to complete successfully prior to moving on the the next step of shutting the system done and creating an AMI or other machine image from the hardened source. With the above change there is no guarantee that the init finished thus leaving the machine without the AIDE baseline database.

Or we need to figure out what exclusions are needed to make the init complete faster. i.e. It may be worthwhile to exclude some directories from the AIDE run like:
/var
/run

What do you think?

jamescassell:
If we want the task to complete successfully, we should just do it synchronously, or otherwise wait for its completion prior to exiting the playbook.  With the old way, there was no guarantee that the init finished.  It was just forcibly killed after 45 seconds, regardless of its completion status.

shepdelacreme:
Doing it synchronously is fine if we can limit the run time. As AIDE is by default it can take way too long to complete the init and its a blocking task. The way it is currently I can guarantee that it finishes before completing the playbook and if it doesn't finish in 45 seconds its killed and the playbook fails (which is also an acceptable state for my use case but not for yours).

I think excluding some dirs like /var and /run should help with the run time and it makes sense from an integrity monitoring standpoint. I think by default AIDE is essentially monitoring everything in dirs like /var/log which it shouldn't since those files are going to be changing constantly.

Also I think I have an idea for making the AIDE init handler a bit better...I just need to test it first.

jamescassell:
As-is, the playbook will not fail because we never poll to check if the tasks completed.

shepdelacreme:
Wait..if the playbook never fails because we don't poll to check if the task completes then why change it? I thought the task was getting killed if it ran longer than 45 seconds?

jamescassell:
The python code that ansible sends to the node itself has a timeout as specified on the async keyword.  It will timeout whether or not ansible ever checks back in with it.

jamescassell:
You can test it with:
```
---
- hosts: localhost
  gather_facts: no
  tasks:
  - command: sleep 600
    async: 45
    poll: 0
```
Then check how long the sleep task actually lasts on the system.  The playbook will exit immediately and never check for task completion.  The sleep process will be killed after 45 seconds.  It will not make it to the 600 seconds specified.

shepdelacreme:
Ok thanks for the background. I did some local testing as well.

So I think I have a handle on an idea for how to do this so it satisfies both. I'll work up a branch and share to see if you agree.

shepdelacreme:
@jamescassell Do you prefer to have the job backgrounded with no time limit at all on the init job or you are fine with having a longer or configurable time limit but still checking on completion within the Ansible playbook?

shepdelacreme:
Closing and using #115 instead.

PR #110, opened 2018-03-27
Implementing RHEL-07-020660

mikerenfro:
Fails on filenames with spaces. I'd fix this in a script by using `find -print0` instead of `find -print`, and then pass everything into `xargs -0`, but not sure how to go about that here.

### DIFF
--- a/tasks/fix-cat2.yml
+++ b/tasks/fix-cat2.yml
@@ -801,6 +801,7 @@
 - name: "MEDIUM | RHEL-07-020660 | PATCH | All files and directories contained in local interactive user home directories must be owned by the owner of the home directory."
   block:
     - name: "MEDIUM | RHEL-07-020660 | AUDIT | All files and directories contained in local interactive user home directories must be owned by the owner of the home directory."
+<<<<<<< HEAD
       # All files in subdirectories
       shell: find {{ item.split(':')[2] }} -mindepth 2 -type f ! -user {{ item.split(':')[0] }}
 <<<<<<< HEAD
@@ -860,18 +861,33 @@
       with_items: "{{rhel_07_020660_audit_3.results}}"
       when: rhel_07_020660_audit_3 | changed
 >>>>>>> 521e1c2... Implementing correct find and xargs commands for everything but home dotfiles.
+=======
+      shell: "{{ find_command_base }} -print -quit"
+      check_mode: no
+      changed_when: rhel_07_020660_audit.stdout != ""
+      register: rhel_07_020660_audit
+      with_items: "{{ rhel7stig_passwd }}"
+      loop_control:
+        label: "{{ rhel7stig_passwd_label }}"
+      when:
+        - this_item.uid >= 1000
+      vars:
+        this_item: "{{ item }}"
 
     - name: "MEDIUM | RHEL-07-020660 | PATCH | All files and directories contained in local interactive user home directories must be owned by the owner of the home directory."
-      # All non-dotfiles in home directory
-      shell: "{{ item.cmd }} -print0 | xargs -r -0 chown {{ item.item.split(':')[0] }}"
-      with_items: "{{rhel_07_020660_audit_2.results}}"
-      when: rhel_07_020660_audit_2 | changed
+      shell: "{{ find_command_base }} -print0 | xargs -r -0 chown {{ this_item.uid }}"
+      with_items: "{{ rhel_07_020660_audit.results }}"
+      loop_control:
+        label: "{{ rhel7stig_passwd_label }}"
+      when:
+        - item | changed
+      vars:
+        this_item: "{{ item.item }}"
+>>>>>>> f3754c8... Adding James' find commands
 
-    - name: "MEDIUM | RHEL-07-020660 | PATCH | All files and directories contained in local interactive user home directories must be owned by the owner of the home directory."
-      # Everything not a file, at any depth
-      shell: "{{ item.cmd }} -print0 | xargs -r -0 chown {{ item.item.split(':')[0] }}"
-      with_items: "{{rhel_07_020660_audit_3.results}}"
-      when: rhel_07_020660_audit_3 | changed
+  vars:
+    find_command_base: 'find "{{ this_item.dir }}" -mindepth 1 \( -path "{{ this_item.dir }}/.*" -not -path "{{ this_item.dir }}/.*/*" -type f \) -o -not -user {{ this_item.uid }}'
+
   when: rhel_07_020660
   tags:
       - RHEL-07-020660
--- a/tasks/fix-cat2.yml
+++ b/tasks/fix-cat2.yml
@@ -801,67 +801,6 @@
 - name: "MEDIUM | RHEL-07-020660 | PATCH | All files and directories contained in local interactive user home directories must be owned by the owner of the home directory."
   block:
     - name: "MEDIUM | RHEL-07-020660 | AUDIT | All files and directories contained in local interactive user home directories must be owned by the owner of the home directory."
-<<<<<<< HEAD
-      # All files in subdirectories
-      shell: find {{ item.split(':')[2] }} -mindepth 2 -type f ! -user {{ item.split(':')[0] }}
-<<<<<<< HEAD
-      with_items: "{{ rhel7stig_passwd_homedir_audit.stdout_lines }}"
-      changed_when: rhel_07_020660_audit_1.stdout != ""
-      register: rhel_07_020660_audit_1
-      check_mode: no
-
-    - name: "MEDIUM | RHEL-07-020660 | AUDIT | All files and directories contained in local interactive user home directories must be owned by the owner of the home directory."
-      # All non-dotfiles in home directory
-      shell: find {{ item.split(':')[2] }} -maxdepth 1 -type f ! -name '.*' ! -user {{ item.split(':')[0] }}
-      with_items: "{{ rhel7stig_passwd_homedir_audit.stdout_lines }}"
-=======
-      with_items: "{{ rhel7stig_passwd_homedir_audit.stdout_lines }}"
-      changed_when: rhel_07_020660_audit_1.stdout != ""
-      register: rhel_07_020660_audit_1
-      check_mode: no
-
-    - name: "MEDIUM | RHEL-07-020660 | AUDIT | All files and directories contained in local interactive user home directories must be owned by the owner of the home directory."
-      # All non-dotfiles in home directory
-      shell: find {{ item.split(':')[2] }} -maxdepth 1 -type f ! -name '.*' ! -user {{ item.split(':')[0] }}
-      with_items: "{{ rhel7stig_passwd_homedir_audit.stdout_lines }}"
->>>>>>> 521e1c2... Implementing correct find and xargs commands for everything but home dotfiles.
-      changed_when: rhel_07_020660_audit_2.stdout != ""
-      register: rhel_07_020660_audit_2
-
-    - name: "MEDIUM | RHEL-07-020660 | AUDIT | All files and directories contained in local interactive user home directories must be owned by the owner of the home directory."
-      # Everything not a file, at any depth
-      shell: find {{ item.split(':')[2] }} ! -type f ! -user {{ item.split(':')[0] }}
-      with_items: "{{ rhel7stig_passwd_homedir_audit.stdout_lines }}"
-      changed_when: rhel_07_020660_audit_3.stdout != ""
-      register: rhel_07_020660_audit_3
-<<<<<<< HEAD
-
-    - name: "MEDIUM | RHEL-07-020660 | PATCH | All files and directories contained in local interactive user home directories must be owned by the owner of the home directory."
-      # All files in subdirectories
-      shell: "{{ item.cmd }} -print0 | xargs -r -0 chown {{ item.item.split(':')[0] }}"
-      with_items: "{{rhel_07_020660_audit_1.results}}"
-      when: rhel_07_020660_audit_1 | changed
-=======
-
-    - name: "MEDIUM | RHEL-07-020660 | PATCH | All files and directories contained in local interactive user home directories must be owned by the owner of the home directory."
-      # All files in subdirectories
-      shell: "{{ item.cmd }} -print0 | xargs -r -0 chown {{ item.item.split(':')[0] }}"
-      with_items: "{{rhel_07_020660_audit_1.results}}"
-      when: rhel_07_020660_audit_1 | changed
-
-    - name: "MEDIUM | RHEL-07-020660 | PATCH | All files and directories contained in local interactive user home directories must be owned by the owner of the home directory."
-      # All non-dotfiles in home directory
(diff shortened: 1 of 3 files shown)


tasks/fix-cat2.yml
```diff
@@ -711,12 +711,21 @@
...
-  changed_when: no
+  block:
+    - name: "MEDIUM | RHEL-07-020660 | AUDIT | All files and directories contained in local interactive user home directories must be owned by the owner of the home directory."
+      command: find {{ item.split(':')[2] }} ! ( -user {{ item.split(':')[0] }} -o -user root )
+      changed_when: no
+      register: rhel_07_020660_audit
+      with_items: "{{ rhel7stig_passwd_homedir_audit.stdout_lines }}"
+
+    - name: "MEDIUM | RHEL-07-020660 | PATCH | All files and directories contained in local interactive user home directories must be owned by the owner of the home directory."
+      #include_tasks: chown_user_file.yml
```
shepdelacreme: Can you remove this commented code if its not needed.

tasks/fix-cat2.yml
```diff
@@ -711,12 +711,21 @@
...
+      command: chown {{ item.item.split(':')[0] }} {{ item.stdout_lines | join(' ') }}
```
shepdelacreme: Can we use a native Ansible module to do this instead of the command module?

mikerenfro: We might, but the problem I ran into was that the `find` module won't let select files based off the owner. Thus the `command` version in the audit task, and we end up with using its `stdout_lines`.

Alternative 1: in the audit task, recurse through each home directory with the `find` module and a register, regardless of owning user. In the patch task, iterate over that register and use the `file` module to set ownership on each file found. The issue there is that I think we'd have to decide that all files in the home directory must be owned by the home directory owner, and not root.

Alternative 2: in the audit task, use the command module like I'm doing now to get a list of files not owned by root or the home directory owner, but don't record those in a register. In the patch task, use the shell module to run the same find command with `-print0` instead of the default `-print`, and pass that over a pipe to `xargs -0`. That'll get us the ability to fix files that have spaces in their names, and won't touch any files already owned by root or the home directory owner.

tasks/fix-cat2.yml
mikerenfro: I can.

tasks/fix-cat2.yml
shepdelacreme: Oh I'm fine using `command: find` to find the files to operate on...the Ansible find module lacks a lot and can be rather slow.

I more meant instead of `command: chown` can we use the Ansible file module?

mikerenfro: Yeah, I think so. Iterate over the find command's stdout_lines for the path attribute of a file task.

shepdelacreme:
Is there a reason for putting the file chown task in a separate yml file?

jamescassell:
I think this would be better remediated by a 'find -exec' command.  json_query should also work.  I'd guess he used the separate file because with_nested doesn't work in this case, and he wanted to do an asymmetric nested loop.

mikerenfro:
Basically. Are we all good with using `find` and `xargs` (faster than `-exec` if there's lots of things to change)?

tasks/fix-cat2.yml
```diff
@@ -711,12 +711,22 @@
       - RHEL-07-020650
 
 - name: "MEDIUM | RHEL-07-020660 | PATCH | All files and directories contained in local interactive user home directories must be owned by the owner of the home directory."
-  command: "true"
-  changed_when: no
+  block:
+    - name: "MEDIUM | RHEL-07-020660 | AUDIT | All files and directories contained in local interactive user home directories must be owned by the owner of the home directory."
+      command: find {{ item.split(':')[2] }} ! ( -user {{ item.split(':')[0] }} -o -user root )
+      changed_when: no
```
jamescassell: I'd say:
```
changed_when:
    - rhel_07_020660_audit.stdout != ""
```

tasks/fix-cat2.yml
```diff
@@ -711,12 +711,22 @@
...
+      register: rhel_07_020660_audit
+      with_items: "{{ rhel7stig_passwd_homedir_audit.stdout_lines }}"
+
+    - name: "MEDIUM | RHEL-07-020660 | PATCH | All files and directories contained in local interactive user home directories must be owned by the owner of the home directory."
```
jamescassell: Fine to change this to a 'find -exec' or 'find | xargs -0'.  Condition its run on:
```
when: rhel_07_020660_audit | changed
```

mikerenfro:
Think I've got this one now. Will apply to the remaining home directory tasks (group and permissions on regular files; owner, group, and permissions on dotfiles) once this has been reviewed.

tasks/fix-cat2.yml
```diff
@@ -711,12 +711,49 @@
       - RHEL-07-020650
 
 - name: "MEDIUM | RHEL-07-020660 | PATCH | All files and directories contained in local interactive user home directories must be owned by the owner of the home directory."
-  command: "true"
-  changed_when: no
+  block:
+    - name: "MEDIUM | RHEL-07-020660 | AUDIT | All files and directories contained in local interactive user home directories must be owned by the owner of the home directory."
+      # All files in subdirectories
+      shell: find {{ item.split(':')[2] }} -mindepth 2 -type f ! -user {{ item.split(':')[0] }}
```
jamescassell: Add "check_mode: no"

mikerenfro:
Hadn't considered a sufficiently clever single find command for all three categories of files. Will see what's possible.

mikerenfro:
@jamescassell @shepdelacreme Looping back around to 020660 again. If this method with the three find commands looks viable, at least as a first attempt, I'll start adapting it to cover 020670, 020680, 020690, 020700, and 020710 (which means we're very close to completing the STIG IDs I need for our internal baseline configuration).

Best case scenario with a sufficiently clever find command, I think we're still looking at two commands, since dotfiles in the home directory are under slightly different rules than regular files and directories, and would need different xargs commands.

jamescassell:
I don't see the need for 3 find commands.  Why can't you do:
```
command: find {{ user_home }} -mindepth 1 -not -user {{ username }}
```
I don't think I understand your different categories of files.

mikerenfro:
Because 020660-020680 only applies to regular files and directories, not dotfiles in the home directory. (covered by covered by 020690-020710).

Normal files and directories are required to be owned by the user, and must be mode 750 or below. Dotfiles in the home directory can be owned by the user or root, and must be mode 740 or below.

If we're going to chown files and directories to the user, we've got to be able to exclude dotfiles in the home directory, but still pick up dotfiles in subdirectories.

If I'm missing something else on your simpler find command, let me know.

jamescassell:
okay, I see what you're saying.  Interestingly, RHEL-07-020660 also implies that any file pointed to by a symlink from a homedir must also be owned by that user, based on their check command.  I'm not sure I'd be comfortable remediating files outside of a homedir, but we can/should warn on them.

For a single command, you should be able to do:
```
command: find {{ user_home }} -mindepth 1 -path {{ user_home }}/.* -o -not -user {{ username }}
```
That should take care of most cases.  To comply with the check content, we'd need to pass '-L' as the first arg to find, but I'm not comfortable with that in a remediation, as I could see a "user" exploiting this by pointing a symlink out of his homedir to a file of which ownership is desired.

jamescassell:
should have been:
```
command: find {{ user_home }} -mindepth 1 -path {{ user_home }}/.* -prune -o -not -user {{ username }}
```

mikerenfro:
Ok. Will test that last one. But will it correctly find a `.emacs` folder in the home directory (subject to 020660), and ignore a `.bashrc` file (subject to 020690)? Here's my script to generate test cases:

```
#!/bin/bash
cd /home/sysadm_user
mkdir -p dir-wrong-owner .dir-wrong-owner2
touch .wrong-owner wrong-owner .wrong-owner2 wrong-owner-2 wrong-owner3
touch .right_owner .right_owner2 right_owner
touch dir-wrong-owner/wrong_owner3 dir-wrong-owner/.wrong_owner4

chown root:root dir-wrong-owner .dir-wrong-owner2
chown 1998:root .wrong-owner wrong-owner .wrong-owner2 wrong-owner-2
chown sysadm_user:root .right_owner2 right_owner
chown root:root wrong-owner3 .right_owner dir-wrong-owner/*

ls -alR /home/sysadm_user

# /home/sysadm_user/:
# total 0
# drwx------. 4 sysadm_user sysadm_user 215 Mar 28 14:25 .
# drwxr-xr-x. 3 root        root         25 Mar 27 17:12 ..
# drwxr-xr-x. 2 root        root         47 Mar 28 14:04 dir-wrong-owner
# drwxr-xr-x. 2 root        root          6 Mar 28 14:14 .dir-wrong-owner2
# -rw-r--r--. 1 root        root          0 Mar 28 13:59 .right_owner
# -rw-r--r--. 1 sysadm_user root          0 Mar 27 10:03 right_owner
# -rw-r--r--. 1 sysadm_user root          0 Mar 28 14:00 .right_owner2
# -rw-r--r--. 1        1999 root          0 Mar 28 13:59 .wrong-owner
# -rw-r--r--. 1        2000 root          0 Mar 27 10:03 wrong-owner
# -rw-r--r--. 1        1998 root          0 Mar 28 13:59 .wrong-owner2
# -rw-r--r--. 1        2001 root          0 Mar 27 15:00 wrong-owner-2
# -rw-r--r--. 1 root        root          0 Mar 28 14:25 wrong-owner3

# /home/sysadm_user/dir-wrong-owner:
# total 0
# drwxr-xr-x. 2 root        root         47 Mar 28 14:04 .
# drwx------. 4 sysadm_user sysadm_user 215 Mar 28 14:25 ..
# -rw-r--r--. 1 root        root          0 Mar 28 14:02 wrong_owner3
# -rw-r--r--. 1 root        root          0 Mar 28 14:04 .wrong_owner4

# /home/sysadm_user/.dir-wrong-owner2:
# total 0
# drwxr-xr-x. 2 root        root          6 Mar 28 14:14 .
# drwx------. 4 sysadm_user sysadm_user 215 Mar 28 14:25 ..
```

jamescassell:
The check command given in the STIG is:
```
ls -lLR /home/smithj
```

That check command won't find you .emacs folder, either, so what I listed above seems compliant with the letter of the STIG.

jamescassell:
The given command in 020690 would find your .emacs folder:
```
 ls -al /home/smithj/.*
```

mikerenfro:
Going to have to think about this one for a while longer:

`find -print0 | xargs -0` is efficient, and won't get tripped up on weird filenames (including spaces).

Anything using `ls` will have to get split out by `stdout_lines` or the equivalent. Individual file tasks could get slow. Sending `ls` output Into a pipe would have to take precautions for weird filenames.

And the `.emacs` folder is subject to 020660, as it's not covered by 020690 (which explicitly applies to files). `.bashrc` is subject to 020690. So any find or related command for 020660 must include top-level dot-folders, exclude top-level dot-files, and include dot-files and dot-folders at all other levels.

mikerenfro:
So I was wrong about a `.emacs` folder at the top level (it's a file). But the `.ssh` folder can take its place in the test cases.

If someone can quickly find a clever set of regexes or similar for find commands to get us below three separate commands, I'm all for it. Otherwise, I'm good to roll in a block of three individual find commands for now and optimize it later.

jamescassell:
I don't see that 020660 leaves an exception for dotfiles, except that the check command the STIG provides won't find dotfiles or dotfolders.  It seems to me that if we exclude dotfiles from here, there should be no problem also excluding dotfolders.  I would include both dotfiles and dotfolders for a 020690 remediation, covering all files and folders between 020660 and 020690.

mikerenfro:
It doesn't have an explicit exception in 020660, no. But 020680 includes

> Files that begin with a "." are excluded from this requirement.

along with the same language in the rule title of

> All files and directories contained in local interactive user home directories

So I'm interpreting that as 020660-020680 apply to everything that's not a top-level dotfile, since 020690-020710 explicitly call out top-level dotfiles with different rules on their ownership and permissions, even though 020660 or 020670 don't contain any exception language similar to 020680.

There's still the possibility that local initialization files as written in 020690-020710 are for any dotfiles at any level, but that requires the authors to have mistakenly used a non-recursive `ls` command to find initialization files in 020690-020710, when they were able to use a recursive `ls` command in 020660-020680, and `find` commands in various other rules.

shepdelacreme:
I think this one can use the /etc/passwd parsing from #130 to make some of the find commands more clear?

mikerenfro:
It can. Will adapt to that in the morning.

jamescassell:
See #137 for a find command that implements what you described above.

mikerenfro:
@shepdelacreme Either I'm blind or Travis is lying about that last build. I only see 'failed=0' on both the main test and the idempotence test. @jamescassell worked in your find commands, I think.

jamescassell:
```
The command "docker exec --tty $container_id yamllint -c /etc/ansible/roles/role_under_test/.yamllint /etc/ansible/roles/role_under_test/" exited with 1.
```

mikerenfro:
So I need to do lint fixes, then? Ok.

jamescassell:
Is there anything covered here not covered by #137 ? The "Files Changed" tab above is not so helpful at this point...

shepdelacreme:
@mikerenfro I'm going to clean up your branch and squash/rebase all these changes so its more clear what is actually changed in this PR.

Also whats the opinion between this and #137 since they both try and implement the same fixes?

mikerenfro:
I'm good to deprecate this in favor of #137. Just wanted to clean up my mess as much as possible. May want to put the `-print0` and `xargs -r -0` in #137, but that's a minor detail.

shepdelacreme:
Ok I cleaned up this branch and squashed a lot of the commits. It was hairy but I think I got everything sorted. I'm not sure how you are rebasing but generally my workflow looks like this.

I have two remotes set on my local git repo:
origin = my fork
upstream = the source repo (MindPointGroup)

```
git fetch upstream
git rebase -i $(git merge-base <my-branch-name> upstream/devel)
```
This will take you through the normal rebase procedure where you can pick/squash commits.
Once you are done with that just run.
```
git rebase upstream/devel
```
Fix any conflicts that crop up. During the fixing of conflicts is usually where I get messed up...especially in large rebases. But generally I just remember to only `git add` or `git rm` after fixing a conflict. No need to `git commit`. 

Then just `git rebase --continue` and keep fixing conflicts until everything is sorted. Or `git rebase --abort` if I've screwed everything up. After your are done you will need to force push. `git push -f`. If you see a bunch of commits show up in the pull request on github that are from another PR that was already merged in then you've done something wrong. You should only see your commits.

To verify on your local branch before force pushing that you don't have extra commits you can review the output of `git log` to make sure there are duplicate commits from others showing up in the log.

FWIW I'm no expert when it comes to git and I'm sure my method above could be much improved by a git master.

shepdelacreme:
Ok closing this and using #137

PR #113, opened 2018-03-27
properly report changes as such

### LINKED ISSUE #112: RHEL-07-010240 makes changes without showing "changed"
RHEL-07-010240 has a task that has "changed_when: no" even though it does make a change.

PR incoming.

jamescassell:
fixes #112

### DIFF
--- a/tasks/fix-cat2.yml
+++ b/tasks/fix-cat2.yml
@@ -287,7 +287,6 @@
 
 - name: "MEDIUM | RHEL-07-010240 | PATCH | Passwords must be restricted to a 24 hours/1 day minimum lifetime."
   command: chage -m 1 {{ item }}
-  changed_when: no
   with_items: "{{ rhel_07_010240_audit.stdout_lines | default([]) }}"
   when: rhel_07_010240
   tags:


jamescassell:
Just pushed the other PR on top of this one because travis was unhappy.

defaults/main.yml
```diff
@@ -69,7 +69,7 @@ rhel_07_010230: true
 rhel_07_010240: true
 rhel_07_010250: true
 # Potential for locking out users
-rhel_07_010260: false
+rhel_07_010260: true
```
shepdelacreme: I'm on the fence about enabling this by default. It changes the known behavior of this role and still has the potential to lock people out of their AWS instances...just not immediately.

jamescassell: Okay, I'm fine merging all but the last commit.  That's why I made it a separate commit.  The role won't lock the root account.  See the awk changes earlier.

shepdelacreme: Ok. If you remove that last commit from this PR I'll merge it in.

jamescassell: ok, done.

jamescassell:
Un-pushed the trailing commit, leaving the default as-is.

I think we should have a policy for which remediation tasks are disabled by default.

shepdelacreme:
Thanks. Agreed on the need for a policy.

jamescassell:
here's the deleted commit for historical interest: https://github.com/jamescassell/RHEL7-STIG/commits/fix112v1

PR #115, opened 2018-03-27
update AIDE init tasks for configurable timeout

shepdelacreme:


shepdelacreme:
@jamescassell Please let me know if this accomplishes what you need. You can set `rhel7stig_aide_init_timeout` to a sufficiently high value and then `rhel7stig_wait_for_aide_init: false` so the playbook won't check status.

tasks/prelim.yml
```diff
@@ -243,11 +243,32 @@
...
+        async: "{{ rhel7stig_aide_init_timeout | default(omit) }}"
+        poll: "{{ (rhel7stig_aide_init_timeout is defined) | ternary(0,omit) }}"
+        register: rhel7stig_aide_init_status
+        when: not rhel7stig_aide_db_status.stat.exists
+
+      - name: "PRELIM | RHEL-07-020030 RHEL-07-020040 RHEL-07-021600 RHEL-07-021610 RHEL-07-021620 | Notify AIDE init handler"
+        command: "true"
+        when:
+          - rhel7stig_aide_init_status.changed
+          - rhel7stig_wait_for_aide_init
```
jamescassell: I'd just move this line to the handler itself.

tasks/prelim.yml
```diff
@@ -243,11 +243,32 @@
...
+        command: /usr/sbin/aide --init -B 'database_out=file:{{ rhel7stig_aide_db_file }}'
+        async: "{{ rhel7stig_aide_init_timeout | default(omit) }}"
+        poll: "{{ (rhel7stig_aide_init_timeout is defined) | ternary(0,omit) }}"
+        register: rhel7stig_aide_init_status
+        when: not rhel7stig_aide_db_status.stat.exists
+
+      - name: "PRELIM | RHEL-07-020030 RHEL-07-020040 RHEL-07-021600 RHEL-07-021610 RHEL-07-021620 | Notify AIDE init handler"
+        command: "true"
+        when:
+          - rhel7stig_aide_init_status.changed
```
jamescassell: Move the notify to the previous task instead.

jamescassell:
After sleeping on it, I think it is probably best to wait to run the init until after all the other STIG remediation tasks are complete.  This is to prevent aide from taking a snapshot of files that will be changed by later steps in the the role.

shepdelacreme:
Noted. I moved it back to be a handler but added in the options to have it run in multiple modes:

- Fire and forget by setting the polling interval to 0
- blocking/sync mode by setting timeout to 0
- "async" mode by setting timeout and polling interval to > 0 (this is the default mode although it operates the same as sync mode

Also noticed that we probably shouldn't be writing directly to /var/lib/aide/aide.db.gz so added logic to write to /var/lib/aide/aide.db.new.gz then move it to /var/lib/aide/aide.db.gz unless you are running the init task in "fire and forget" mode then it just writes it directly.

shepdelacreme:
Once this is merged I think we should add notify to the tasks that edit /etc/aide.conf as well so that a new baseline db is generated with the updated settings.

tasks/prelim.yml
```diff
@@ -249,26 +249,20 @@
...
-        register: rhel7stig_aide_init_status
-        when: not rhel7stig_aide_db_status.stat.exists
-
-      - name: "PRELIM | RHEL-07-020030 RHEL-07-020040 RHEL-07-021600 RHEL-07-021610 RHEL-07-021620 | Notify AIDE init handler"
-        command: "true"
-        when:
-          - rhel7stig_aide_init_status.changed
-          - rhel7stig_wait_for_aide_init
-        notify: init aide status
+      - name: "PRELIM | RHEL-07-020030 RHEL-07-020040 RHEL-07-021600 RHEL-07-021610 RHEL-07-021620 | Set AIDE database output file"
```
jamescassell: There's no need for this task.  Always write to the temp file.  Worst case, put it in defaults or vars.

handlers/main.yml
```diff
@@ -23,13 +23,20 @@
...
+  async: "{{ rhel7stig_aide_init_timeout | default(omit) }}"
+  poll: "{{ rhel7stig_aide_init_interval | default(omit) }}"
+  when: not rhel7stig_aide_db_status.stat.exists
+  notify: move aide db
+  tags:
+    - aide
+
+- name: move aide db
+  command: "mv {{ rhel7stig_aide_db_output_file }} {{ rhel7stig_aide_db_file }}"
+  when:
```
jamescassell: Do this unconditionally.  AIDE init will continue writing to the file originally opened, even if the file changes fs location underneath it.  This is how the RHEL6-STIG role works.  (That role still has the 'killed after 45 seconds' problem, though.)

handlers/main.yml
```diff
@@ -23,13 +23,20 @@
...
 
-- name: init aide status
-  async_status:
-    jid: "{{ rhel7stig_aide_init_status.ansible_job_id }}"
-  register: rhel7stig_aide_init_job_result
-  until: rhel7stig_aide_init_job_result.finished
-  retries: 60
-  delay: 5
+- name: init aide
+  command: /usr/sbin/aide --init -B 'database_out=file:{{ rhel7stig_aide_db_output_file }}'
```
jamescassell: Probably want to move the aide init to after any other handlers that make filesystem changes.

tasks/prelim.yml
```diff
@@ -249,26 +249,20 @@
...
             name: aide
             state: present
+        notify: init aide
 
       - name: "PRELIM | RHEL-07-020030 RHEL-07-020040 RHEL-07-021600 RHEL-07-021610 RHEL-07-021620 | Check for existing AIDE database"
         stat:
           path: "{{ rhel7stig_aide_db_file }}"
         register: rhel7stig_aide_db_status
         check_mode: no
+        changed_when: not rhel7stig_aide_db_status.stat.exists
```
jamescassell: Perfect!

jamescassell:
Is it desirable to have 'aide --check' return clean after running this role and a reboot?

handlers/main.yml
shepdelacreme: Hmm is this as simple as moving these handlers to the end of the file?

jamescassell: handlers are always executed in the order specified in the handlers file, regardless of the order they were notified.

handlers/main.yml
```diff
@@ -56,3 +39,17 @@
...
+  poll: "{{ rhel7stig_aide_init_interval | default(omit) }}"
+  when: not rhel7stig_aide_db_status.stat.exists
+  notify: move aide db
+  tags:
+    - aide
+
+- name: move aide db
+  command: "mv {{ rhel7stig_aide_temp_db_file }} {{ rhel7stig_aide_db_file }}"
+  tags:
+    - aide
```
jamescassell: Please have valid lines.

handlers/main.yml
```diff
@@ -56,3 +39,17 @@
...
   command: auditctl -f {{ rhel7stig_auditd_failure_flag }}
   failed_when: no
+
+- name: init aide
+  command: /usr/sbin/aide --init -B 'database_out=file:{{ rhel7stig_aide_temp_db_file }}'
+  async: "{{ rhel7stig_aide_init_timeout | default(omit) }}"
+  poll: "{{ rhel7stig_aide_init_interval | default(omit) }}"
+  when: not rhel7stig_aide_db_status.stat.exists
+  notify: move aide db
+  tags:
```
jamescassell: Is there a need for tags in the handlers file?

handlers/main.yml
```diff
@@ -56,3 +39,17 @@
...
+    - aide
+
+- name: move aide db
+  command: "mv {{ rhel7stig_aide_temp_db_file }} {{ rhel7stig_aide_db_file }}"
+  tags:
```
jamescassell: handlers behave as if they had
```
tags:
- always
```

handlers/main.yml
shepdelacreme: We should probably move the reboot handler to the bottom of the file then.

handlers/main.yml
shepdelacreme: There used to be a need but it looks like they fixed that.

handlers/main.yml
jamescassell: True, though I don't think we use the reboot handler.

shepdelacreme:
Just tested running `aide --check` after a reboot and it comes back unclean but only because of the its noticing the ansible async job file is missing.
```
---------------------------------------------------
Removed files:
---------------------------------------------------

removed: /root/.ansible_async/228982313278.15010.tmp
```

jamescassell:
If we get rid of async and always do it synchronously, or with the true fire-and-forget mode I posted we won't get that.

shepdelacreme:
I'm fine doing it synchronously but I need a way to ensure the task finishes before shutting the box down.

jamescassell:
Presumably it will usually finish before the default SSH timeout of 10 minutes.  Otherwise, the play will fail and you will know.  In my experience, even if it doesn't finish within the SSH timeout period, the aide init still finishes successfully, eventually.

shepdelacreme:
Yes most of the time. However if I'm running this with packer or other similar image builders it is going to shut the machine down immediately after the playbook finishes. There is the potential for the task to just not finish unless the shutdown command would wait for the aide init process to finish before shutting down.

shepdelacreme:
How about this then? sync/blocking or fire and forget only. Fire and forget uses your shell nohup command...accomplished with two different handlers set using a var.

tasks/prelim.yml
```diff
@@ -243,11 +243,21 @@
...
+            name: aide
+            state: present
+        notify: init aide
+
+      - name: "PRELIM | RHEL-07-020030 RHEL-07-020040 RHEL-07-021600 RHEL-07-021610 RHEL-07-021620 | Check for existing AIDE database"
+        stat:
+          path: "{{ rhel7stig_aide_db_file }}"
+        register: rhel7stig_aide_db_status
+        check_mode: no
+        changed_when: not rhel7stig_aide_db_status.stat.exists
```
jamescassell: Since we're conditionally notifying the handler here, we don't need the "when" on the handler itself.

handlers/main.yml
```diff
@@ -46,3 +39,16 @@
 - name: update running audit failure mode
   command: auditctl -f {{ rhel7stig_auditd_failure_flag }}
   failed_when: no
+
+- name: init aide and wait
+  command: /usr/sbin/aide --init -B 'database_out=file:{{ rhel7stig_aide_temp_db_file }}'
+  when: not rhel7stig_aide_db_status.stat.exists
```
jamescassell: we can get rid of this line now.

handlers/main.yml
```diff
@@ -46,3 +39,16 @@
...
+  notify: move aide db
+
+- name: init aide
+  shell: nohup /usr/sbin/aide --init -B 'database_out=file:{{ rhel7stig_aide_temp_db_file }}' > /dev/null &
+  when: not rhel7stig_aide_db_status.stat.exists
```
jamescassell: and this line

jamescassell:
I like your new approach.

tasks/prelim.yml
shepdelacreme: Good catch

shepdelacreme:
The thought of backing up an old/existing DB has crossed my mind. I guess it somewhat depends how people use this role. I generally just run it on a new/clean box at the beginning of the build process. Are people running this role on existing machines? On a regular schedule, like every month or something?

I suppose we could have a var:
`rhel7stig_force_init_aide_db: true`
That would for AIDE init regardless of the outcome of the `stat` task.

Then add a task after the `stat` task that moves the existing DB file to `/var/lib/aide/aide.db.{{ datatime }}.gz` or something. However I don't want to bother adding all that logic if no one really cares and is fine just overwriting the existing DB.

jamescassell:
I'm fine over-writing the DB.  We only do it when something changed that would necessitate re-creating it anyway.

shepdelacreme:
So just a force or overwrite db variable? I think I can just drop that on the stat task. Should be easy.

jamescassell:
maybe a default "rhel7stig_allow_init_aide_db: true" then change the handler "when" to be "when: rhel7stig_allow_init_aide_db"

jamescassell:
Or better yet, just put it on the mv task and call it rhel7stig_allow_overwrite_aide_db

tasks/fix-cat2.yml
```diff
@@ -897,6 +897,7 @@
         regexp: '([A-Z]+ = .*)(sha256\+sha512)(.*)'
         replace: '\1sha512\3'
       register: rhel_07_021620_sha256_512_result
+      notify: "{{ rhel7stig_aide_handler }}"
       failed_when:
         - rhel_07_021620_sha256_512_result | failed
         - rhel_07_021620_sha256_512_result.rc != 257
```
jamescassell: While we're making changes here, we can get rid of the custom failed_when since this is now tied to the PRELIM task.  Could be a separate PR, though.

tasks/fix-cat2.yml
```diff
@@ -907,11 +908,13 @@
         regexp: '([A-Z]+ = .*)(sha256)(.*)'
         replace: '\1sha512\3'
       register: rhel_07_021620_sha256_result
+      notify: "{{ rhel7stig_aide_handler }}"
       failed_when:
```
jamescassell: same here

tasks/prelim.yml
```diff
@@ -249,7 +249,7 @@
         yum:
             name: aide
             state: present
-        notify: init aide
+        notify: "{{ rhel7stig_aide_handler }}"
```
jamescassell: Good.

handlers/main.yml
```diff
@@ -42,13 +42,12 @@
...
   notify: move aide db
 
 - name: init aide
   shell: nohup /usr/sbin/aide --init -B 'database_out=file:{{ rhel7stig_aide_temp_db_file }}' > /dev/null &
-  when: not rhel7stig_aide_db_status.stat.exists
   notify: move aide db
 
 - name: move aide db
   command: "mv {{ rhel7stig_aide_temp_db_file }} {{ rhel7stig_aide_db_file }}"
+  when: not rhel7stig_aide_db_status.stat.exists or rhel7stig_overwrite_aide_db
```
jamescassell: Looks good.

tasks/fix-cat2.yml
shepdelacreme: There isn't a stat task for /etc/aide.conf in prelim yet so the failed_when is technically still needed for now. I'll add that in a later PR I think.

jamescassell: later PR is fine.  /etc/aide.conf is provided by the aide package, so it's fine to error out if it's not there.

jamescassell:
Awesome! Ship it!

PR #116, opened 2018-03-27
Implementing RHEL-07-020270

mikerenfro:
May have misinterpreted what a good AUDIT task would be.


jamescassell:
I'd like to see a better remediation for this, but I can't think of what "better" would mean.  At least, it would probably involve excluding default system accounts other than 'games'  These generally have uid < 299...  It seems like the "right" way to do this is to require a whitelist, not a blacklist.  Unfortunately, that would make it difficult to enabe this by default.

PR #117, opened 2018-03-27
Implementing RHEL-07-020900.

mikerenfro:


tasks/fix-cat2.yml
```diff
@@ -774,13 +774,21 @@
...
       - notimplemented
 
-- name: "MEDIUM | RHEL-07-020900 | PATCH | All system device files must be correctly labeled to prevent unauthorized modification."
-  command: "true"
-  changed_when: no
+- name: "MEDIUM | RHEL-07-020900 | AUDIT | All system device files must be correctly labeled to prevent unauthorized modification."
+  block:
+    - name: "MEDIUM | RHEL-07-020900 | AUDIT | All system device files must be correctly labeled to prevent unauthorized modification."
+      command: find {{ ansible_mounts | to_json | from_json | json_query("[?starts_with(device, '/dev/')].mount") | join(' ') }} ( -context *:device_t:* -o -context *:unlabeled_t:* ) ( -type c -o -type b ) -printf '%p %Z\n'
+      when: ansible_selinux
```
jamescassell: You should add the condition to the block below.

tasks/fix-cat2.yml
```diff
@@ -774,13 +774,21 @@
...
+      changed_when: no
```
jamescassell: you'll also want "check_mode: no"

PR #118, opened 2018-03-28
Implementing 032010

mikerenfro:
Can remove the `savlog` command, but all sites would have to set it to their preferred command. If someone can tell me what the CLAM equivalent would be, I can edit the default.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -282,7 +282,10 @@ rhel7stig_av_package:
 # RHEL-07-032010
 # How to check if your definitions are up to date. Passed to an Ansible
 # shell task, and that task's stdout_lines will be printed in a debug task.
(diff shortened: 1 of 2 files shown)


shepdelacreme:
From the STIG:

> If "McAfee VirusScan Enterprise for Linux" is active on the system, check the dates of the virus definition files with the following command:
>
>`# ls -al /opt/NAI/LinuxShield/engine/dat/*.dat`
><need output>
>
>If the virus definition files have dates older than seven days from the current date, this is a finding.
>
>If "clamav" is active on the system, check the dates of the virus database with the following commands:
>
>`# grep -I databasedirectory /etc/clamav.conf`
>DatabaseDirectory /var/lib/clamav
>
>`# ls -al /var/lib/clamav/*.cvd`
>-rwxr-xr-x 1 root root 149156 Mar 5 2011 daily.cvd

PR #128, opened 2018-03-29
Fixing 122

mikerenfro:
Only setting permissions if the home directory's stat results indicate group `w` or other `rwx`.

### DIFF
--- a/tasks/fix-cat2.yml
+++ b/tasks/fix-cat2.yml
@@ -713,11 +713,25 @@
       - RHEL-07-020620
 
 - name: "MEDIUM | RHEL-07-020630 | PATCH | All local interactive user home directories must have mode 0750 or less permissive."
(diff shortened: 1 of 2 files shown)


tasks/fix-cat2.yml
```diff
@@ -713,11 +713,25 @@
...
+      file:
+        path: "{{ item.stat.path }}"
+        mode: "{{ rhel7stig_homedir_mode }}"
+        state: directory
+      with_items: "{{ rhel_07_020630_audit.results }}"
+      when:
+        - item.stat.wgrp or
+          item.stat.roth or
+          item.stat.woth or
+          item.stat.xoth
```
jamescassell: I'd move these conditions to a changed_when above, and make this conditional on "changed"

PR #129, opened 2018-03-29
Fix 121

mikerenfro:
Leaves the default cron syslog location alone. If there's a clever way to ensure that we don't modify an otherwise `cron.*` line pointing to a different file, I'm all for it.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -465,7 +465,7 @@ rhel7stig_audit_disk_size_query: "[?mount=='{{ rhel7stig_audit_part }}'].size_to
 rhel7stig_auditd_mail_acct: root
 
 # RHEL-07-020630
-rhel7stig_homedir_mode: 0700
+rhel7stig_homedir_mode: 0750
 
 # RHEL-07-031000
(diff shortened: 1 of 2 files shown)


defaults/main.yml
```diff
@@ -465,7 +465,7 @@ rhel7stig_audit_disk_size_query: "[?mount=='{{ rhel7stig_audit_part }}'].size_to
 rhel7stig_auditd_mail_acct: root
 
 # RHEL-07-020630
-rhel7stig_homedir_mode: 0750
+rhel7stig_homedir_mode: 0700
```
jamescassell: looks like this change sneaked into the PR.

tasks/fix-cat2.yml
```diff
@@ -876,8 +876,7 @@
   lineinfile:
     path: /etc/rsyslog.conf
     state: present
-    regexp: 'cron.\*'
```
jamescassell: perhaps:
```
regexp: '^cron\.\*[ \t]+/var/log/cron$'
```

PR #130, opened 2018-04-02
parse /etc/passwd into data structure

jamescassell:
This improves legibility of the code.

I took the structure names from the "ansible_user_*" fact names.


jamescassell:
I've got another PR that depends on this one, which I'll send once this is merged.

shepdelacreme:
Thanks for this! I like the readability this adds!

PR #134, opened 2018-04-02
Implementing RHEL-07-021000

mikerenfro:
If `nosuid` is missing from `/home` mount options, and `/home` is a separate filesystem, add `nosuid` to the current mount options.

This may change the `/home` options even if it's NFS-mounted, which falls under 021020. But there's no distinction made in 021000 for the type of filesystem. Since the only way to avoid a finding on 021000 (if applicable) is to have `/home` be a part of `/`, I'm reading this as "add the `nosuid` option to `/home`, regardless of its physical location".


jamescassell:
TASK [R7S : MEDIUM | RHEL-07-021000 | PATCH | File systems that contain user home directories must be mounted to prevent files with the setuid and setgid bit set from being executed.] **
fatal: [host]: FAILED! => {
    "changed": false
}

MSG:

missing required arguments: state

PR #135, opened 2018-04-03
Implementing 021020 and 021021

mikerenfro:
Will probably need to be rebased if James' local_mounts consolidation goes in first. Doesn't exactly match the STIG check, since it looks through `/etc/fstab`, and this looks through active mounts.


shepdelacreme:
Thanks! LGTM. Needs a rebase.

PR #139, opened 2018-04-06
Workaround for benchmarks

jamescassell:
These changes make both scap benchmarks happier: scap-security-guide scan with oscap and DISA content V1R2 with spawarscc.


shepdelacreme:
Nice! I like this method for those situations that the benchmarks are being obtuse.

PR #145, opened 2018-04-18
Proof-of-concept: handling of disruptive and complex tasks

jamescassell:
This starts implementation of the proposal I made in https://github.com/MindPointGroup/RHEL7-STIG/issues/119#issuecomment-381254864

This implements the complexity-high and distruption-high concepts, vars, and tags, with an example of each.

Proposal quoted:

> Here's my proposal.
> The scap-security-guide project labels their remediation tasks with 3 categories: severity, complexity, and disruption.  It seems like a good approach to me:
> 
> - I suggest we tag items requiring manual review as `complexity-high` and add a conditional on `rhel7stig_complexity_high: default(no)`.
> - I suggest we tag items that can break a system as `disruption-high` and add a conditional on `rhel7stig_disruption_high: default(no)`.
> 
> We can then add later `disruption-medium` and `disruption-low` for tasks that are respectively less likely to break something.  Similarly, later add `complexity-medium` and `complexity-low` for tasks depending on how involved the remediation is.  (Specifically, I'm thinking that if, e.g., a remediation requires multiple tasks to succeed, it would get `complexity-medium`.)
> 
> Edit: @shepdelacreme yes, I'd also like to add something like `rhel7stig_audit_complex` and `rhel7stig_audit_disruptive` as an available hint to the user that something was not remediated.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,7 @@ script:
     # Test role idempotence.
     - idempotence_log=$(mktemp)
 
-    - docker exec --tty $container_id ansible-playbook /etc/ansible/roles/role_under_test/tests/travis.yml --extra-vars "@/etc/ansible/roles/role_under_test/tests/extra_vars_travis.yml" -vv | tee -a $idempotence_log
+    - docker exec --tty $container_id ansible-playbook /etc/ansible/roles/role_under_test/tests/travis.yml --extra-vars "@/etc/ansible/roles/role_under_test/tests/extra_vars_travis.yml" --extra-vars "@/etc/ansible/roles/role_under_test/tests/extra_vars_travis_idempotence.yml" -vv | tee -a $idempotence_log
     - >
       tail $idempotence_log
       | grep -q "changed=0.*failed=0"
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -49,7 +49,7 @@ rhel_07_020210: true
 rhel_07_020220: true
 rhel_07_020230: true
 # Not an automated task
-rhel_07_020250: false
+rhel_07_020250: true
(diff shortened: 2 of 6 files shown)


jamescassell:
I believe these commits are ready to merge.  I've opened the PR now to get feedback on the general approach before I spend time applying the concept to the rest of the role.

shepdelacreme:
Thanks @jamescassell Sorry I am just now getting around to looking at this in depth. 

My only issue is that it does add some amount of complexity/confusion. I find it hard to look at these changes and understand exactly what this role is going to do. 

For instance: With `rhel7stig_complexity_high: no` and `rhel7stig_audit_complex: yes` will the role repeatedly show changed on task RHEL-07-020250 if it is not a supported version?

In the instance of RHEL-07-020250 that may make sense since it is warning a user about an out of date OS version. But it may be confusing to users that are not familiar with these settings why it is continuously reporting changed. It also intentionally breaks the idempotence of the role...which is a major change.

I'm not opposed to this but I would like to hear some more feedback. Also how many tasks/items you think these would be applied to?

jamescassell:
You are correct that running the role multiple times with complexity_high: no and audit_complex: yes will result in a "changed" result every time if the item is non-compliant.  I'd be fine making the default audit_complex: no to keep the "idempotence" of the role.  (in reality, the role is idempotent even with the change since no change is made in any case regardless of the audit_complex status; only reports "changed" to get the user's attention.)

As far as what other tasks would get the complexity-high tag: that would be things that currently just report to the user via "debug."  Those would gain a changed_when item to flag when an out-of-compliance state is detected.  disruption-high would apply to any tasks we decide to disable by default to avoid breaking machines.

PR #149, opened 2018-05-31
cmd: fails to execute /usr/sbin/postconf

drewgwallace:
Fresh CentOS minimal updated to 7.5 as target.

### DIFF
--- a/tasks/fix-cat2.yml
+++ b/tasks/fix-cat2.yml
@@ -2122,14 +2122,14 @@
         register: rhel_07_040680_rpm_audit
 
       - name: "MEDIUM | RHEL-07-040680 | AUDIT | The system must be configured to prevent unrestricted mail relaying."
-        shell: "$(whereis postconf | cut -d ' ' -f2) -n smtpd_client_restrictions"
+        shell: "/usr/sbin/postconf -n smtpd_client_restrictions"
         check_mode: no
         changed_when: no
         register: rhel_07_040680_postconf_audit
         when: rhel_07_040680_rpm_audit.rc == 0
 
       - name: "MEDIUM | RHEL-07-040680 | PATCH | The system must be configured to prevent unrestricted mail relaying."
-        shell: "$(whereis postconf | cut -d ' ' -f2) -e 'smtpd_client_restrictions=permit_mynetworks, reject'"
+        shell: "/usr/sbin/postconf -e 'smtpd_client_restrictions=permit_mynetworks, reject'"
         when:
             - rhel_07_040680_rpm_audit.rc == 0
             - rhel_07_040680_postconf_audit.stdout != 'smtpd_client_restrictions = permit_mynetworks, reject'
--- a/tasks/fix-cat2.yml
+++ b/tasks/fix-cat2.yml
@@ -2122,14 +2122,14 @@
         register: rhel_07_040680_rpm_audit
 
       - name: "MEDIUM | RHEL-07-040680 | AUDIT | The system must be configured to prevent unrestricted mail relaying."
-        command: "postconf -n smtpd_client_restrictions"
+        shell: "$(whereis postconf | cut -d ' ' -f2) -n smtpd_client_restrictions"
         check_mode: no
         changed_when: no
         register: rhel_07_040680_postconf_audit
         when: rhel_07_040680_rpm_audit.rc == 0
 
       - name: "MEDIUM | RHEL-07-040680 | PATCH | The system must be configured to prevent unrestricted mail relaying."
-        command: "postconf -e 'smtpd_client_restrictions=permit_mynetworks, reject'"
+        shell: "$(whereis postconf | cut -d ' ' -f2) -e 'smtpd_client_restrictions=permit_mynetworks, reject'"
         when:
             - rhel_07_040680_rpm_audit.rc == 0
             - rhel_07_040680_postconf_audit.stdout != 'smtpd_client_restrictions = permit_mynetworks, reject'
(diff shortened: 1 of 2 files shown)


shepdelacreme:
Any idea what changed in 7.5 that it couldn't execute postconf?

jamescassell:
I'll try to reproduce and pinpoint any issue as soon as I have time.

drewgwallace:
This is with Ansible 2.5.3
I tried with a 7.4 system as well and receive the same error:

    fatal: [centos74.xxxx]: FAILED! => {
    "changed": false,
    "cmd": "postconf -n smtpd_client_restrictions",
    "invocation": {
        "module_args": {
            "_raw_params": "postconf -n smtpd_client_restrictions",
            "_uses_shell": false,
            "chdir": null,
            "creates": null,
            "executable": null,
            "removes": null,
            "stdin": null,
            "warn": true
        }
    },
    "msg": "[Errno 2] No such file or directory",
    "rc": 2

jamescassell:
I've not yet tried to reproduce the issue, but it seems that your path environment variable is broken.

jamescassell:
I've seen a broken path environment variable with use of the dzdo become method

drewgwallace:
It does appear to be a path configuration issue with sshd
From /etc/ssh/sshd_config
`# This sshd was compiled with PATH=/usr/local/bin:/usr/bin`
A non-interactive check of the $PATH variable confirms this. Is this not sshd compilation options not the same in RHEL?

jamescassell:
I could not reproduce this error on a clean minimal install of RHEL or CentOS.

drewgwallace:
Thank you for looking into this. Could you post the stdout of these tasks run against the working target system?
```
- shell: cat /etc/ssh/sshd_config | grep "# This sshd"
- command: echo $PATH
- command: whereis postconf
```
Also, what is your become method?

jamescassell:
@drewgwallace per your request:
```
# This sshd was compiled with PATH=/usr/local/bin:/usr/bin
/sbin:/bin:/usr/sbin:/usr/bin
postconf: /usr/sbin/postconf /usr/share/man/man1/postconf.1.gz /usr/share/man/man5/postconf.5.gz
```

Using sudo.

drewgwallace:
Regarding this task:
`- command: echo $PATH`

- become_method: sudo
    `"stdout": "/sbin:/bin:/usr/sbin:/usr/bin"`

- become_method: su
    `"stdout": "/usr/local/bin:/usr/bin"`

Executing a remote SSH command I get the noninteractive $PATH. Unsure what Ansible is doing with the sudo become_method.
```
ssh user@1.2.3.4 'sudo -S  echo $PATH'
user@1.2.3.4's password:
[sudo] password for user: 1234
/usr/local/bin:/usr/bin
```

jamescassell:
run your playbook with '-vvv' and ansible will show you exactly what it's running behind the scenes.

drewgwallace:
Any remote execution of an Ansible generated command.py as an unprivileged user or root will result in a PATH set from `/etc/ssh/sshd_config`, the `/etc/sudoers` file changes this.
```
cat /etc/sudoers | grep "Defaults    secure_path"
Defaults    secure_path = /sbin:/bin:/usr/sbin:/usr/bin
```

It leads me to believe these `command:` tasks utilizing binaries from `sbin/` directories will not function without `become_method: sudo` or an explicit path e.g. `/usr/sbin/postconf`.

Example excerpts:
<pre>ssh -t user@1.2.3.4 '/bin/sh -c "<b>sudo</b> /usr/bin/python /home/user/command.py"'
        "stdout": "/sbin:/bin:/usr/sbin:/usr/bin"</pre>


<pre>ssh -t user@1.2.3.4 '/bin/sh -c "    /usr/bin/python /home/user/command.py"'
        "stdout": "/usr/local/bin:/usr/bin"</pre>

What do you think?

jamescassell:
@drewgwallace in ansible-2.4, you can set:
ansible_become_flags="--login"

That will give you the correct path.  They seem to have broken that facility in ansible-2.5, so probably worth opening a bug report.

jamescassell:
@shepdelacreme I'd be inclined to close this PR, and perhaps document that either become_method=sudo is required, or
1. PATH be manually set to a sane value at the playbook level or,
2. set ansible_become_flags="--login" to get root's proper PATH (currently broken on ansible-2.5 for me, works for 2.4)

shepdelacreme:
@jamescassell What about using the explicit path to postconf?  i.e. `/usr/sbin/postconf`. I think we ended up switching some other tasks use explicit paths as well. I like it because it is specific about what we are trying to execute.

jamescassell:
I wouldn't be opposed to using the absolute path. There are certainly other tasks that might need a change, though.

drewgwallace:
These are the task changes I've located that permit the default role to execute completely.

tasks/fix-cat2.yml
```diff
@@ -2122,14 +2122,14 @@
...
-        command: "postconf -n smtpd_client_restrictions"
+        shell: "/usr/sbin/postconf -n smtpd_client_restrictions"
         check_mode: no
         changed_when: no
         register: rhel_07_040680_postconf_audit
         when: rhel_07_040680_rpm_audit.rc == 0
 
       - name: "MEDIUM | RHEL-07-040680 | PATCH | The system must be configured to prevent unrestricted mail relaying."
-        command: "postconf -e 'smtpd_client_restrictions=permit_mynetworks, reject'"
+        shell: "/usr/sbin/postconf -e 'smtpd_client_restrictions=permit_mynetworks, reject'"
```
jamescassell: likewise

PR #150, opened 2018-06-01
Change yum group to x11 id value

drewgwallace:
There is variation in the groupname, change to id referral to remove variation
    `yum grouplist hidden id | grep X`
      `X Window System (x11)`
      `Legacy UNIX Compatibility (legacy-unix)`
      `Legacy X Window System Compatibility (legacy-x)`

### DIFF
--- a/tasks/fix-cat2.yml
+++ b/tasks/fix-cat2.yml
@@ -2154,10 +2154,10 @@
   tags:
       - RHEL-07-040720
 
-- name: "MEDIUM | RHEL-07-040730 | PATCH | An X Windows display manager must not be installed unless approved."
+- name: "MEDIUM | RHEL-07-040730 | PATCH | An X Window display manager must not be installed unless approved."
   yum:
       name:
-          - "@X Windows System"
+          - "@x11"
           - xorg-x11-server-common
       state: absent
   when:


jamescassell:
Looks good to me, as long as it works on both RHEK and CentOS.

jamescassell:
looks like it should work on CentOS and RHEL, but I haven't actually tested.

jamescassell:
@drewgwallace can you unpush the last 2 commits?

jamescassell:
Once the 2 spurious commits are unpushed, I'm okay with merging this.

@shepdelacreme Another item I noticed in testing this PR: we remove @ x11 before we set the option to automatically remove dependencies, so we end up with unneeded deps of @ x11 remaining installed on the system even though they are no longer required.  For the moment, this can be worked around by first running the role with '--skip-tags=RHEL-07-040730' to get the "auto remove deps" option set then with '--tags=RHEL-07-040730' to make sure @ x11 is gone from the system.

drewgwallace:
Okay that should do it.

shepdelacreme:
@jamescassell hmm ok. Which task is the "auto remove deps" option set in? It might be worth setting that in prelim since there are a fair number of other package removal tasks.

jamescassell:
clean_requirements_on_remove in /etc/yum.conf

PR #152, opened 2018-06-05
Complex rework

jamescassell:
followup to #145
also fixes some fatal errors we ran into with assert tasks that should have been debug tasks

### DIFF
--- a/tasks/fix-cat2.yml
+++ b/tasks/fix-cat2.yml
@@ -712,7 +712,6 @@
             msg: "Manual intervention is required -- missing owner on items in {{ item.item.mount }}: {{ item.stdout_lines | join(', ') }}"
         changed_when:
             - rhel7stig_audit_complex
-            - "{{ item.stdout_lines | length }} == 0"
         when:
             - item.stdout_lines is defined
             - item.stdout_lines | length > 0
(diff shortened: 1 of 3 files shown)


tasks/fix-cat2.yml
```diff
@@ -1015,12 +1023,16 @@
       - name: "MEDIUM | RHEL-07-020900 | AUDIT | All system device files must be correctly labeled to prevent unauthorized modification."
         debug:
             msg: "{{ rhel_07_020900_audit.stdout_lines }}"
+        changed_when:
+            - rhel7stig_audit_complex
```
shepdelacreme: Should there be a conditional here to only show changed when `rhel_07_020900_audit.stdout_lines` contains something?

tasks/fix-cat2.yml
```diff
@@ -1603,10 +1615,15 @@
       - name: "MEDIUM | RHEL-07-032010 | AUDIT | The system must update the DoD-approved virus scan program every seven days or more frequently."
         debug:
             msg: "{{ rhel_07_032010_audit.stdout_lines }}"
+        changed_when:
+            - rhel7stig_audit_complex
```
shepdelacreme: Should there be a conditional here to only show changed when rhel_07_032010_audit.stdout_lines contains something?

tasks/fix-cat2.yml
jamescassell: agreed

tasks/fix-cat2.yml
jamescassell: likewise

jamescassell:
@shepdelacreme good catch on the idempotence.  I've pushed one more commit that should fix that.

shepdelacreme:
LGTM! Thanks!

PR #157, opened 2018-07-27
Acs stigs

jhultz:
Added in ssh intervals and changed ansible to reflect developer needs. 

Disabled some Stigs by default because they relabel the entire system which isn't optimal for development with custom packages that have permissions changed after install.


defaults/main.yml
```diff
@@ -31,8 +31,8 @@ rhel7stig_workaround_for_ssg_benchmark: true
 # PLEASE NOTE: These work in coordination with the cat1, cat2, cat3 group variables. You must enable an entire group
 # in order for the variables below to take effect.
 # CAT 1 rules
-rhel_07_010010: true
-rhel_07_010020: true
+rhel_07_010010: false
```
shepdelacreme: I don't think the consensus would be to have these disabled by default in the role. If you want to override the application of some of these fixes for specific machines then you can set these variables on a per-machine basis using group or host vars for the dev machines.

meta/main.yml
```diff
@@ -2,7 +2,7 @@
 galaxy_info:
     author: "Sam Doran, Josh Springer, Daniel Shepherd, James Cassell, Mike Renfro"
     description: "Apply the DISA RHEL 7 STIG"
-    company: "MindPoint Group"
+    company:
```
shepdelacreme: Needs a rebase

vars/main.yml
```diff
@@ -1,5 +1,5 @@
 ---
-rhel7stig_min_ansible_version: 2.4
+rhel7stig_min_ansible_version: 2.5
```
shepdelacreme: Is there a problem with the role and Ansible 2.4?

vars/main.yml
```diff
@@ -32,3 +32,22 @@ rhel7stig_re_qp_key_end: (?:" *)
...
 rhel7stig_re_qp_insert: "{{ insert | default(not (append | default(true))) }}"
+
+#Developers said the STIG lockdowns made ssh too difficult as the CAT2 finding says we must disconnect after 10 minutes of inactivity on ssh and console sessions. This will not be done on production/integration environments, but if teams want it in their sandbox, modify the following values:
+# rhel_07_040320_ssh_keepalive_interval:
+# rhel_07_040160_console_keepalive_interval:
+#Both values are in seconds (60 seconds * 10 = 600)
+             
+#The following will set the ssh timeout to 1 day:
+#RHEL-07-040320 | All network connections associated with SSH traffic must terminate at the end of the session or after 10 minutes of inactivity, except to fulfill documented and validated mission requirements
+rhel_07_040320_ssh_keepalive_interval: 28800
```
shepdelacreme: This is a good add but should be placed in `defaults/main.yml` and set to the default setting that meets the STIG `600` and then overridden in your group or host vars files, not here in `vars/main.yml`.

vars/main.yml
```diff
@@ -32,3 +32,22 @@ rhel7stig_re_qp_key_end: (?:" *)
...
+             
+#RHEL-07-040160 | All network connections associated with a communication session must be terminated at the  the session or after 10 minutes of inactivity from the user at a command prompt, except to fulfill documented and validated mission requirements.
+rhel_07_040160_console_keepalive_interval: 28800
```
shepdelacreme: There is already a var to control this setting in `defaults/main.yml` at https://github.com/sscpac/RHEL7-STIG/blob/8819962f2d44334a876148944419166d98a73240/defaults/main.yml#L442-L444

Should override this setting in your group/host vars as needed.

vars/main.yml
```diff
@@ -32,3 +32,22 @@ rhel7stig_re_qp_key_end: (?:" *)
...
+         
+#Requiring a bootloader password is required under the STIG. If you are on a AWS/C2S environment, this is not applicable and you need to disable this value in a vars.yml file
+#Set these to false to disable it.
+rhel_07_010480: false
+rhel_07_010490: false
```
shepdelacreme: Should override this setting in your group/host vars as needed.

defaults/main.yml
jhultz: On a vanilla install I haven't run into any issues, but when working in dev environments these 2 are the root cause of either system breaks or in cases of multi cluster installs I watched this take over 24 hours of run time to get through because of custom software utilizing ansible after RPM install to change ownership of files/directories.

douglasrapp: @jhultz - But our custom RPMs are the exception. Generally speaking this role would be run on a fresh RHEL installation, where that problem wouldn't be encountered. We can disable this on our own runs if executing after our RPMs get installed.

vars/main.yml
jhultz: This was just to keep up with the latest so my users were not running 2.4 instead of 2.5 when 2.6 is already out.

dirtyharrycallahan: Will you also address the deprecation warnings such as "Using tests as filters is deprecated"

jamescassell: @dirtyharrycallahan see #144. Are there outlier cases that I missed?

dirtyharrycallahan: I only use the role on a system w/o internet access so I won't know until next week when I perform my transfers.

shepdelacreme: We don't need to change this in the role at this time. If it is to make sure local users are not running older versions of Ansible I think that is better handled as a site specific thing. i.e. site policy about versions of software packages or overriding this variable locally. 

We would update this value to a newer version if we started using features, modules, etc that only exist in the newer version of Ansible. Hence it is the minimum required Ansible version.

meta/main.yml
```diff
@@ -2,7 +2,7 @@
 galaxy_info:
     author: "Sam Doran, Josh Springer, Daniel Shepherd, James Cassell, Mike Renfro"
     description: "Apply the DISA RHEL 7 STIG"
-    company: "MindPoint Group"
+    company: "SSCPAC"
```
shepdelacreme: Please back out this change.

defaults/main.yml
```diff
@@ -136,7 +136,7 @@ rhel_07_021000: true
 rhel_07_021010: true
 rhel_07_021020: true
 rhel_07_021021: true
-rhel_07_021030: true
+rhel_07_021030: false
```
shepdelacreme: Why has this rule be disabled by default?

meta/main.yml
jhultz: The change has been reverted

meta/main.yml
```diff
@@ -2,7 +2,7 @@
 galaxy_info:
     author: "Sam Doran, Josh Springer, Daniel Shepherd, James Cassell, Mike Renfro"
     description: "Apply the DISA RHEL 7 STIG"
-    company: "SSCPAC"
+    company:
```
shepdelacreme: This line needs to read:
`company: "MindPoint Group"`

as per this commit https://github.com/MindPointGroup/RHEL7-STIG/commit/0523e4dba87b599f9034d5455967810b315495fb It was a fix so that our import of this role on Ansible Galaxy worked correctly.

shepdelacreme:
Closing - superseded by #167

PR #161, opened 2018-08-02
minimal install compatibility

### LINKED ISSUE #158: role attempts to remediate SSH even if not installed
If openssh-server is not installed, none of the SSH remediations need to be performed.  Currently, the role fails out in this case.

### LINKED ISSUE #160: disable wifi fails without NetworkManager installed
PR incoming.

jamescassell:
These fixes improve compatibility with a minimal install.

fixes #158
fixes #160


tasks/prelim.yml
```diff
@@ -56,6 +56,19 @@
   tags:
       - always
 
+- name: ensure cronie is available
+  yum:
+      name: cronie
+  when:
+      - rhel_07_020030 or
+        rhel_07_020040 or
+        false
```
shepdelacreme: What is the purpose of the `or false`?

shepdelacreme:
LGTM. Just one comment.

tasks/prelim.yml
jamescassell: Allows reordering the above items without worrying about which one doesn't have an or at the end, and makes it so adding new items doesn't require adding or to an existing item

jamescassell: But in this case, the list is quite short and unlikely to change frequently. I'm happy to remove it if that is preferred.

shepdelacreme: I'm fine leaving it in I just wasn't sure what the purpose was lol. Now it makes sense.

PR #165, opened 2018-08-02
improve and fix RPM verify

### LINKED ISSUE #151: RHEL-07-010010 PATCH gets tripped up by missing files
Hello,

I don't know if this is a common problem and it's hard to understand this for me because I do not have in-depth knowledge of RedHat / CentOS / general rpm-based distros but on a fresh install of CentOS 7 from `CentOS-7-x86_64-Minimal-1804.iso` there are lots of missing package files that result in `rpm --setperms` exiting with an error return code.

Example:

```
[root@localhost ~]# rpm --setperms $(rpm -qf /usr/sbin/netreport)
chmod: cannot access ‘/etc/sysconfig/kvm’: No such file or directory
```

This causes RHEL-07-010010 to fail a lot.

Is there something wrong with my CentOS installation or is missing package files normal?

jamescassell:
fixes #151 

- more friendly output
- don't error out
- speed improvements
- squash bug with setuid files

### DIFF
--- a/tasks/fix-cat1.yml
+++ b/tasks/fix-cat1.yml
@@ -21,7 +21,7 @@
 
 - block:
       - name: "HIGH | RHEL-07-010020 | AUDIT | The cryptographic hash of system files and commands must match vendor values."
-        shell: 'rpm -Va | grep ''^..5'' | grep -E ''/.?bin/'''
+        shell: 'rpm -Va | grep ''^..5'' | grep -E ''/(.?bin|libexec)/'''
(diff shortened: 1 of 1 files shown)


shepdelacreme:
LGTM! Thanks for splitting out the commits for these changes like this btw! Made it very easy to review each change.

jamescassell:
Also makes troubleshooting easier down the road when you're trying to figure out why a certain thing is a certain way.

PR #166, opened 2018-08-06
feature: unprivileged container support

jamescassell:
after these changes, the role can be run on unprivileged containers running systemd (such as the rhel7-init container) by setting:
```
rhel7stig_system_is_container: yes
```

The role can also be run on unprivileged non-systemd-running containers by additionally setting:
```
rhel7stig_system_is_chroot: yes
```


shepdelacreme:
This is great. With this we should also be able to move away from using the `skip_for_travis` var since, for the most part, those items are skipped because travis is running in a container.

PR #169, opened 2018-08-15
New testing

defionscode:
WIP for molecule testing


jamescassell:
This seems to have removed the yamllint check from Travis CI.  Can we add it back?

PR #170, opened 2018-08-15
Travis + Molecule

defionscode:


molecule/aws/molecule.yml
```diff
@@ -0,0 +1,59 @@
+---
+driver:
+  name: ec2
+lint:
+  name: yamllint
+  options:
```
jamescassell: #169 removed the .yamllint file that had this content:
```yaml
extends: default
rules:
    indentation:
        spaces: 4
    truthy: disable
    line-length: disable
    # todo: it would be nice to limit lines to 132
    # line-length:
    #     max: 132
    #     allow-non-breakable-inline-mappings: true
```

shepdelacreme: I tried this it doesn't cause molecule to lint the ansible playbooks it only ends up linting things in the molecule directory.

PR #176, opened 2018-10-13
STIG v2r1 updates - deleted rules

shepdelacreme:
Removes:
  - RHEL-07-020070: removed (YAY, no more repo_gpgcheck that always fails)
  - RHEL-07-030600: removed (audit /var/log/tallylog)
  - RHEL-07-030730: removed (audit sudoedit)
  - RHEL-07-030850: removed (audit rmmod, should modify RHEL-07-030840 first)
  - RHEL-07-030860: removed (audit modprobe, should modify RHEL-07-030840 first)
  - RHEL-07-032010: removed (weekly virus cvd/dat updates)
  - RHEL-07-010062: removed (prevent a user from overriding the screensaver lock-enabled setting)

Modifies:
  - RHEL-07-030840: "insmod" -> "kmod"


shepdelacreme:
I'm pretty sure its the new way we are testing. Since there are private build vars used in Travis we can't expose them when PRs are built/tested from forks.

jamescassell:
That's unfortunate...

shepdelacreme:
Yeah...there isn't really a good way to get around it if you want to have credentials and what not used during the job. Open to suggestions/etc.

PR #180, opened 2018-10-16
STIG v2r1 updates - batch 2

shepdelacreme:
- RHEL-07-010150: "dcredit" -> "ocredit" in the fix text
- RHEL-07-010200: reference both password-auth and system-auth instead of system-auth-ac
- RHEL-07-010270: references both password-auth and system-auth instead of system-auth-ac, "pam_unix.so"->"pam_pwhistory.so"
- RHEL-07-010320: references both password-auth and system-auth instead of system-auth-ac
  - "unlock_time=604800" -> "fail_interval=900 unlock_time=900"
- RHEL-07-010330: references both password-auth and system-auth instead of password-auth-ac and system-auth-ac
  - "unlock_time=604800" -> "unlock_time=900"
- RHEL-07-010350: non-compliant configs allowed in commented-out lines
- RHEL-07-010430: if the required config is commented, it is a finding
- RHEL-07-020110: stop autofs in addition to disable

- Also made some updates so that the pamd changes in 270, 320, and 330 accept updated values and are idempotent (with a helper)

### DIFF
--- a/tasks/fix-cat2.yml
+++ b/tasks/fix-cat2.yml
@@ -339,10 +339,6 @@
             new_type: password
             new_control: requisite
             new_module_path: pam_pwhistory.so
-            module_arguments:
-              - use_authtok
-              - remember={{ rhel7stig_pam_pwhistory.remember|default(5) }}
-              - retry={{ rhel7stig_pam_pwhistory.retries|default(3) }}
         with_items:
             - "system-auth"
             - "password-auth"
@@ -410,7 +406,6 @@
             new_type: auth
             new_control: required
             new_module_path: pam_faillock.so
-            module_arguments: "preauth silent audit deny={{ rhel7stig_pam_faillock.attempts }}{{ (rhel7stig_pam_faillock.fail_for_root) | ternary(' even_deny_root ',' ') }}fail_interval={{ rhel7stig_pam_faillock.interval }} unlock_time={{ rhel7stig_pam_faillock.unlock_time }}"
         with_items:
             - "system-auth"
             - "password-auth"
@@ -451,7 +446,6 @@
             new_type: auth
             new_control: "[default=die]"
             new_module_path: pam_faillock.so
-            module_arguments: "authfail audit deny={{ rhel7stig_pam_faillock.attempts }}{{ (rhel7stig_pam_faillock.fail_for_root) | ternary(' even_deny_root ',' ') }}fail_interval={{ rhel7stig_pam_faillock.interval }} unlock_time={{ rhel7stig_pam_faillock.unlock_time }}"
         with_items:
             - "system-auth"
             - "password-auth"
--- a/tasks/fix-cat2.yml
+++ b/tasks/fix-cat2.yml
@@ -498,7 +498,7 @@
       - RHEL-07-010330
       - pamd
 
-- name: "MEDIUM | RHEL-07-010340 | PATCH | Users must provide a password for privilege escalation."
+- name: "MEDIUM | RHEL-07-010340 | PATCH | The Red Hat Enterprise Linux operating system must be configured so that users must provide a password for privilege escalation."
   replace:
       path: "{{ item }}"
       regexp: '^([^#].*)NOPASSWD(.*)'
@@ -512,7 +512,7 @@
       - RHEL-07-010340
       - sudoers
 
-- name: "MEDIUM | RHEL-07-010350 | PATCH | Users must re-authenticate for privilege escalation."
+- name: "MEDIUM | RHEL-07-010350 | PATCH | The Red Hat Enterprise Linux operating system must be configured so that users must re-authenticate for privilege escalation."
   replace:
       path: "{{ item }}"
       regexp: '^([^#].*)!authenticate(.*)'
--- a/tasks/fix-cat2.yml
+++ b/tasks/fix-cat2.yml
@@ -524,7 +524,7 @@
       - RHEL-07-010350
       - sudoers
 
-- name: "MEDIUM | RHEL-07-010430 | PATCH | The delay between logon prompts following a failed console logon attempt must be at least four seconds."
+- name: "MEDIUM | RHEL-07-010430 | PATCH | The Red Hat Enterprise Linux operating system must be configured so that the delay between logon prompts following a failed console logon attempt is at least four seconds."
   lineinfile:
       dest: /etc/login.defs
       regexp: ^#?FAIL_DELAY
--- a/tasks/fix-cat2.yml
+++ b/tasks/fix-cat2.yml
@@ -343,7 +343,7 @@
             - "system-auth"
             - "password-auth"
 
-      # Temporary audit check to determine if the following pamd module task needs to be run since the module is not idempotent
+      # TODO: Remove temporary audit check to determine if the following pamd module task needs to be run since the module is not idempotent
       - name: "MEDIUM | RHEL-07-010270 | AUDIT | Check for existing password history reuse settings"
         command: "grep -iE '^password\\s+requisite\\s+pam_pwhistory.so\\s+use_authtok\\s+remember={{ rhel7stig_pam_pwhistory.remember|default(5) }}\\s+retry={{ rhel7stig_pam_pwhistory.retries|default(3) }}$' /etc/pam.d/{{ item }}"
         check_mode: no
@@ -354,6 +354,7 @@
             - "system-auth"
             - "password-auth"
 
+      # TODO: Once the pamd module is fixed (either args_present or updated) then remove the previous grep task and update the following.
       - name: "MEDIUM | RHEL-07-010270 | PATCH | Ensure pam_pwhistory module arguments are set"
         pamd:
             name: "{{ item.item }}"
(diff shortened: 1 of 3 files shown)


shepdelacreme:
Tests are going to fail until we figure out the game plan for testing PRs from forks but these changes all pass locally.

tasks/fix-cat2.yml
```diff
@@ -325,14 +327,51 @@
...
+        failed_when: rhel_07_010270_audit.rc > 1
+        register: rhel_07_010270_audit
+        with_items:
+            - "system-auth"
+            - "password-auth"
+
+      - name: "MEDIUM | RHEL-07-010270 | PATCH | Ensure pam_pwhistory module arguments are set"
+        pamd:
+            name: "{{ item.item }}"
+            state: updated
```
jamescassell: The `pamd` documentation is very unclear.  The `updated` state will replace the type,control,module_path with new_type,new_control,new_module_path.  I haven't tested, but I'd guess that only as a side effect does the set of module_arguments get updated with this task.  I'd make this `args_present`
```suggestion
            state: args_present
```

tasks/fix-cat2.yml
```diff
@@ -325,14 +327,51 @@
...
+        pamd:
+            name: "{{ item }}"
+            state: before
+            type: password
+            control: sufficient
+            module_path: pam_unix.so
+            new_type: password
+            new_control: requisite
+            new_module_path: pam_pwhistory.so
+            module_arguments:
```
jamescassell: I'd just drop this and allow the next task to take care of it.

tasks/fix-cat2.yml
```diff
@@ -325,14 +327,51 @@
...
+              - use_authtok
+              - remember={{ rhel7stig_pam_pwhistory.remember|default(5) }}
+              - retry={{ rhel7stig_pam_pwhistory.retries|default(3) }}
+        with_items:
+            - "system-auth"
+            - "password-auth"
+
+      # Temporary audit check to determine if the following pamd module task needs to be run since the module is not idempotent
+      - name: "MEDIUM | RHEL-07-010270 | AUDIT | Check for existing password history reuse settings"
```
jamescassell: I'd drop this task entirely.

tasks/fix-cat2.yml
```diff
@@ -325,14 +327,51 @@
...
+            name: "{{ item.item }}"
+            state: updated
+            type: password
+            control: requisite
+            module_path: pam_pwhistory.so
+            module_arguments:
+              - use_authtok
+              - remember={{ rhel7stig_pam_pwhistory.remember|default(5) }}
+              - retry={{ rhel7stig_pam_pwhistory.retries|default(3) }}
+        with_items: "{{rhel_07_010270_audit.results}}"
```
jamescassell: after dropping the "temporary audit check", change this to:
```suggestion
        with_items:
            - system-auth
            - password-auth
```

tasks/fix-cat2.yml
```diff
@@ -325,14 +327,51 @@
...
+        when: item.rc == 1
```
```suggestion
```

tasks/fix-cat2.yml
```diff
@@ -170,7 +170,7 @@
       - RHEL-07-010140
       - pwquality
 
-- name: "MEDIUM | RHEL-07-010150 | PATCH | When passwords are changed or new passwords are assigned, the new password must contain at least one special character."
+- name: "MEDIUM | RHEL-07-010150 | PATCH | The Red Hat Enterprise Linux operating system must be configured so that when passwords are changed or new passwords are established, the new password must contain at least one special character."
```
jamescassell: Seems unrelated to the other changes, but no problem with the updated itself.

tasks/fix-cat2.yml
```diff
@@ -376,7 +415,33 @@
             - "system-auth"
             - "password-auth"
 
-      - name: "MEDIUM | RHEL-07-010330 | PATCH | If three unsuccessful logon attempts within 15 minutes occur the associated account must be locked."
+      # Temporary audit check to determine if the following pamd module task needs to be run since the module is not idempotent
+      - name: |
+              "MEDIUM | RHEL-07-010320 | AUDIT | Check for existing account lockout settings"
+              "MEDIUM | RHEL-07-010330 | AUDIT | Check for existing account lockout settings"
+        command: "grep -iE '^auth\\s+required\\s+pam_faillock.so\\s+preauth\\s+silent\\s+audit\\s+deny={{ rhel7stig_pam_faillock.attempts }}{{ (rhel7stig_pam_faillock.fail_for_root) | ternary(' even_deny_root ',' ') }}fail_interval={{ rhel7stig_pam_faillock.interval }}\\s+unlock_time={{ rhel7stig_pam_faillock.unlock_time }}$' /etc/pam.d/{{ item }}"
```
jamescassell: drop

tasks/fix-cat2.yml
```diff
@@ -376,7 +415,33 @@
...
+        with_items:
+            - "system-auth"
+            - "password-auth"
+
+      - name: |
+              "MEDIUM | RHEL-07-010320 | PATCH | Accounts on the Red Hat Enterprise Linux operating system that are subject to three unsuccessful logon attempts within 15 minutes must be locked for the maximum configurable period."
+              "MEDIUM | RHEL-07-010330 | PATCH | The Red Hat Enterprise Linux operating system must lock the associated account after three unsuccessful root logon attempts are made within a 15-minute period."
+        pamd:
+            name: "{{ item.item }}"
+            state: updated
```
jamescassell: args_present

tasks/fix-cat2.yml
```diff
@@ -391,7 +456,29 @@
             - "system-auth"
             - "password-auth"
 
-      - name: "MEDIUM | RHEL-07-010330 | PATCH | If three unsuccessful logon attempts within 15 minutes occur the associated account must be locked."
+      # Temporary audit check to determine if the following pamd module task needs to be run since the module is not idempotent
+      - name: "MEDIUM | RHEL-07-010330 | AUDIT | Check for existing account lockout settings"
+        command: "grep -iE '^auth\\s+\\[default=die\\]\\s+pam_faillock.so\\s+authfail\\s+audit\\s+deny={{ rhel7stig_pam_faillock.attempts }}{{ (rhel7stig_pam_faillock.fail_for_root) | ternary(' even_deny_root ',' ') }}fail_interval={{ rhel7stig_pam_faillock.interval }}\\s+unlock_time={{ rhel7stig_pam_faillock.unlock_time }}$' /etc/pam.d/{{ item }}"
```
jamescassell: drop

tasks/fix-cat2.yml
```diff
@@ -391,7 +456,29 @@
...
+        failed_when: rhel_07_010330_authfail_audit.rc > 1
+        register: rhel_07_010330_authfail_audit
+        with_items:
+            - "system-auth"
+            - "password-auth"
+
+      - name: "MEDIUM | RHEL-07-010330 | PATCH | The Red Hat Enterprise Linux operating system must lock the associated account after three unsuccessful root logon attempts are made within a 15-minute period."
+        pamd:
+            name: "{{ item.item }}"
+            state: updated
```
jamescassell: args_present

tasks/fix-cat2.yml
jamescassell: That's a bug in the pamd module, then...

jamescassell:
I'll see if I can fiddle with this tomorrow morning if it hasn't been worked out

tasks/fix-cat2.yml
shepdelacreme: So I think I need to keep this task until the pamd module can be fixed and then I have 1 of 2 options. Keep using the `updated` task below to make the change (this works correctly with argument value changes) OR I can move to `args_present` but if I do that I'll need to add ANOTHER task to remove all the args before I re-add them.

The current way seems like the least hacky way until we can get the pamd module fixed.

shepdelacreme:
I'm going to rebase this and make the `module_arguments` change. But I don't think it makes sense to move to `args_present` since its broken and the only way I see for this to work for now is keeping the `grep` task and using `updated`.

Let me know what you find out when you have a chance to fiddle.

jamescassell:
Also would be great to add integration tests Upstream for the ansible pamd module to catch regressions prior to release.

shepdelacreme:
Yup I've been submitting issues about this stuff. Also talked today about getting some time to work on contributing to/fixing the module.

Made the updates and added the TODO comments.

jamescassell:
I seem to recall that updating the module arguments worked correctly as long as each argument was its own list item and not a space separated string. I'm not in front of a computer right now but I'll test as soon as I have a chance.

defionscode:
I think it might make sense to leverage GH projects to track this sort of stuff. Will be easier to remember than relying on TODO in code. I think it's fine to keep TODO comments in the YAML but as a sole source of TODOs, I always forget about them in regular day-to-day work/iteration.

I'm going to merge this in. The bigger picture priority is passing SCAP checks and I think we can achieve that with this. We can always tweak behind-the-scenes implementation (ie when pamd module is patched to not suck) and it will still be the same user-experience and outcome to anyone who uses this role.

jamescassell:
Since this has already been merged, I'll fiddle with these pamd items if/when I get around to it.

PR #181, opened 2018-10-16
Docker updates

defionscode:
Improved compatability with Docker images. Partitioning requirements dont really apply in a container context so those have be pulled out. Also sshd related checks are now only checked when rhel7stig_ssh_required is set to true (which is the default); containers with sshd is almost always a bad practice to begin with. Also, installs openssh packages when ssh is required, the docs currently have said it installs ssh when request, but that was not previously true.


tasks/fix-cat2.yml
```diff
@@ -2061,10 +2061,10 @@
       name: "{{ rhel7stig_firewall_service }}"
       state: "{{ rhel7stig_service_started }}"
       enabled: yes
-  # XXX: fix this.  The above task should work just fine in a chroot, but
-  # fails for some reason when the chroot is also a container.
-  ignore_errors: "{{ rhel7stig_system_is_chroot and rhel7stig_system_is_container }}"
-  when: rhel_07_040520
+  when:
+      - rhel_07_040520
+      - not rhel7stig_system_is_chroot
```
jamescassell: This will prevent the task from running in a postinstall ks script.  Will firewalld run in a systemd container?

tasks/fix-cat3.yml
```diff
@@ -84,10 +87,10 @@
...
       masked: no
       state: "{{ rhel7stig_service_started }}"
-  # XXX: fix this.  The above task should work just fine in a chroot, but
-  # fails for some reason when the chroot is also a container.
-  ignore_errors: "{{ rhel7stig_system_is_chroot and rhel7stig_system_is_container }}"
-  when: rhel_07_021340
+  when:
+      - rhel_07_021340
+      - not rhel7stig_system_is_container
+      - not rhel7stig_system_is_chroot
```
jamescassell: likewise

tasks/prelim.yml
```diff
@@ -190,3 +190,27 @@
...
+      name: python-passlib
+  when:
+      - inventory_hostname == 'localhost' or inventory_hostname == '127.0.0.1'
+      - rhel_07_020030 or rhel_07_010480
+
+- name: "PRELIM | RHEL-07-020210 RHEL-07-020220| Install SELinux related dependencies"
+  yum:
+      name: "{{ item }}"
+  with_items:
+      - selinux-policy
```
jamescassell: would you ever install selinux-policy without either selinux-policy-targeted or selinux-policy-mls?  I think I've broken a system with only installing selinux-policy, but I could be mis-remembering...

tasks/prelim.yml
```diff
@@ -190,3 +190,27 @@
...
+      - libselinux-python
+  when:
+      - rhel_07_020210 or rhel_07_020220
+
+- name: "PRELIM | Install SSH"
+  yum:
+      name: "{{ item }}"
+  with_items:
+      - openssh-server
+      - openssh-clients
```
jamescassell: openssh-clients is no longer required by the STIG as of V2R1.

tasks/fix-cat2.yml
jamescassell: It was only 'ignore_errors' when both system_is_chroot and system_is_container were set.  I.e., A non-systemd container.  The other cases worked fine.

local.yml
```diff
@@ -2,5 +2,9 @@
 - hosts: localhost
   connection: local
   become: yes
+  vars:
+    ansible_python_interpreter: "{{ ansible_playbook_python }}"
   roles:
-      - RHEL7-STIG
+    - role: RHEL7-STIG
+      rhel7stig_system_is_container: "{{ is_container | default(false) }}"
+      rhel7stig_ssh_required: "{{ is_container == false  }}"
```
jamescassell: Does this work? (i.e., does comparing an undefined value to false return true?)

tasks/fix-cat2.yml
defionscode: Ah that was a mistake on my part, I can change it to `and` i just wanted to skip the task rather than ignore errors. Agreed on DISA, NIST already has some guidance but it's quite generic. As far as detecting capabilities at run time, my knowledge falls short, I do not know how you know what caps the container itself has unless your checking from the docker host itself to be able to do `getcap` or `netcap` or similar. Ideas?

jamescassell: `capsh --print` might be a good start... not sure, and unfortunately I don't have time now to find out.

defionscode: ah, TIL, that would work. I just tested.

defionscode:
Made the updates, I skipped doing a CAP_NET_ADMIN check for now since it is not a real problem yet, in that no user reports having firewalld issues with containers atm, if it comes up we can deal with it then.

RHEL7-STIG
```diff
@@ -0,0 +1 @@
+.
```
jamescassell: didn't we already do this in https://github.com/MindPointGroup/RHEL7-STIG/commit/19af3fd6b8b67e53e948df0245058ab4413ef06b ?  We don't actually need it because we can use "{{ playbook_dir }}" to reference the role in the current directory.

local.yml
```diff
@@ -2,5 +2,10 @@
 - hosts: localhost
   connection: local
   become: yes
+  vars:
+    ansible_python_interpreter: "{{ ansible_playbook_python }}"
+    is_container: false
   roles:
-      - RHEL7-STIG
+    - role: RHEL7-STIG
```
jamescassell:
```suggestion
    - role: "{{ playbook_dir }}"
```

tasks/fix-cat2.yml
```diff
@@ -2145,10 +2145,9 @@
       name: "{{ rhel7stig_firewall_service }}"
       state: "{{ rhel7stig_service_started }}"
       enabled: yes
-  # XXX: fix this.  The above task should work just fine in a chroot, but
-  # fails for some reason when the chroot is also a container.
-  ignore_errors: "{{ rhel7stig_system_is_chroot and rhel7stig_system_is_container }}"
-  when: rhel_07_040520
+  when:
+      - rhel_07_040520
+      - not rhel7stig_system_is_chroot and not rhel7stig_system_is_container
```
jamescassell: "NOT BOTH" vs "NEITHER"

```suggestion
      - not (rhel7stig_system_is_chroot and rhel7stig_system_is_container)
```

tasks/fix-cat3.yml
```diff
@@ -84,10 +87,9 @@
       enabled: yes
       masked: no
       state: "{{ rhel7stig_service_started }}"
-  # XXX: fix this.  The above task should work just fine in a chroot, but
-  # fails for some reason when the chroot is also a container.
-  ignore_errors: "{{ rhel7stig_system_is_chroot and rhel7stig_system_is_container }}"
-  when: rhel_07_021340
+  when:
+      - rhel_07_021340
+      - not rhel7stig_system_is_container and not rhel7stig_system_is_chroot
```
jamescassell: same as above

tasks/prelim.yml
```diff
@@ -190,3 +190,28 @@
...
+      - rhel_07_020030 or rhel_07_010480
+
+- name: "PRELIM | RHEL-07-020210 RHEL-07-020220| Install SELinux related dependencies"
+  yum:
+      name: "{{ item }}"
+  with_items:
+      - selinux-policy
+      - libselinux-python
+      - selinux-policy-targeted
+      - selinux-policy-mls
```
jamescassell: does it work to install both selinux-policy-targeted and selinux-policy-mls?

RHEL7-STIG
shepdelacreme: That is my fault I removed it in my testing update not realizing it was still used by local.yml

tasks/prelim.yml
defionscode: yep, it works, you still have to set the policy explicitly independent of the package being installed/available IIRC

local.yml
```diff
@@ -2,5 +2,10 @@
 - hosts: localhost
   connection: local
   become: yes
+  vars:
+    ansible_python_interpreter: "{{ ansible_playbook_python }}"
+    is_container: false
   roles:
-      - RHEL7-STIG
+    - role: "{{ playbook_dir }}"
```
jamescassell: Needs 2 more spaces of indentation per ansible-lockdown coding standard.

local.yml
```diff
@@ -2,5 +2,10 @@
 - hosts: localhost
   connection: local
   become: yes
+  vars:
+    ansible_python_interpreter: "{{ ansible_playbook_python }}"
```
jamescassell: 2 more spaces of indentation

tasks/prelim.yml
jamescassell: This is pulled in as a dep of the other policy packages below.

tasks/prelim.yml
```diff
@@ -190,3 +190,28 @@
...
+  when:
+      - rhel_07_020210 or rhel_07_020220
+
+- name: "PRELIM | Install SSH"
+  yum:
+      name: "{{ item }}"
```
jamescassell: Can just collapse this... no need to loop on a single item.

defionscode:
Man im suckin on this one, updated.

local.yml
```diff
@@ -2,5 +2,11 @@
...
```
jamescassell: Do you remember why `ansible_python_interpreter` was added here?  It's set automatically by ansible exactly as you have done here with all versions of ansible back to 2.4.  (2.4 does not template ansible_x_interpreter values, so this playbook fails on 2.4.)

PR #183, opened 2018-10-17
add new cat 3 rules

shepdelacreme:
RHEL-07-021022 - nodev
RHEL-07-021023 - nosuid
RHEL-07-021024 - noexec /dev/shm mount options

### DIFF
--- a/tasks/fix-cat3.yml
+++ b/tasks/fix-cat3.yml
@@ -32,6 +32,34 @@
       - complexity-high
       - passwd
 
+- debug:
+    msg: "{{ dev_shm_mount }}"
+  vars:
+      dev_shm_mount: "{{ ansible_mounts | json_query('[?mount == `/dev/shm`] | [0]') }}"
+  tags:
+    - always
+
+- name: |
+        "LOW | RHEL-07-021022 | PATCH | The Red Hat Enterprise Linux operating system must mount /dev/shm with the nodev option."
+        "LOW | RHEL-07-021023 | PATCH | The Red Hat Enterprise Linux operating system must mount /dev/shm with the nosuid option."
+        "LOW | RHEL-07-021024 | PATCH | The Red Hat Enterprise Linux operating system must mount /dev/shm with the noexec option."
+  mount:
+      path: /dev/shm
+      state: mounted
+      src: "{{ dev_shm_mount.device|default('tmpfs') }}"
+      fstype: "{{ dev_shm_mount.fstype|default('tmpfs') }}"
+      opts: "{{ dev_shm_mount.options|default('defaults') }},nodev,nosuid,noexec"
+  when:
+      - rhel_07_021022
+      - rhel_07_021023
+      - rhel_07_021024
+  vars:
+      dev_shm_mount: "{{ ansible_mounts | json_query('[?mount == `/dev/shm`] | [0]') }}"
+  tags:
+      - RHEL-07-021022
+      - RHEL-07-021023
+      - RHEL-07-021024
+
 - name: "LOW | RHEL-07-021310 | PATCH | A separate file system must be used for user home directories (such as /home or an equivalent)."
   debug:
       msg: "WARNING: /home is not mounted on a separate partition"
--- a/tasks/fix-cat3.yml
+++ b/tasks/fix-cat3.yml
@@ -32,13 +32,6 @@
       - complexity-high
       - passwd
 
-- debug:
-    msg: "{{ dev_shm_mount }}"
-  vars:
-      dev_shm_mount: "{{ ansible_mounts | json_query('[?mount == `/dev/shm`] | [0]') }}"
-  tags:
-    - always
-
 - name: |
         "LOW | RHEL-07-021022 | PATCH | The Red Hat Enterprise Linux operating system must mount /dev/shm with the nodev option."
         "LOW | RHEL-07-021023 | PATCH | The Red Hat Enterprise Linux operating system must mount /dev/shm with the nosuid option."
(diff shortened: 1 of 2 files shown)


tasks/fix-cat3.yml
```diff
@@ -32,6 +32,34 @@
...
+- name: |
+        "LOW | RHEL-07-021022 | PATCH | The Red Hat Enterprise Linux operating system must mount /dev/shm with the nodev option."
+        "LOW | RHEL-07-021023 | PATCH | The Red Hat Enterprise Linux operating system must mount /dev/shm with the nosuid option."
+        "LOW | RHEL-07-021024 | PATCH | The Red Hat Enterprise Linux operating system must mount /dev/shm with the noexec option."
+  mount:
+      path: /dev/shm
+      state: mounted
+      src: "{{ dev_shm_mount.device|default('tmpfs') }}"
+      fstype: "{{ dev_shm_mount.fstype|default('tmpfs') }}"
+      opts: "{{ dev_shm_mount.options|default('defaults') }},nodev,nosuid,noexec"
```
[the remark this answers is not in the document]
shepdelacreme: Yes the mount module is fully idempotent and also handles the remount when its needed.

I was following what we did here: https://github.com/MindPointGroup/RHEL7-CIS/blob/be56adb2f0b4f30b284a8bbc481092d121d1b8f2/tasks/section1.yml#L372

I did go down the path of making this split so that you could do some, all, or none of these settings but is anyone really going to have a need to apply 1 of these options and not the others? If I split these out it would require at a minimum several extra tasks. I can certainly do that but was avoiding the complexity that adds.

shepdelacreme:
Here is some more info for this set of tasks.

By default in RHEL7 `/dev/shm` has no entry in `/etc/fstab` but it is mounted with the nosuid and nodev options. i.e.
```
grep shm /proc/mounts
tmpfs /dev/shm tmpfs rw,seclabel,nosuid,nodev 0 0
```
However the STIG specifically checks `/etc/fstab` so even by default you would fail those on an audit unless you add the line to fstab. If we split out all 3 and make them optional independently than that means you could disable the nodev and nosuid ones and essentially make your system less secure than it is by default.

I can do something like this which is largely the same as what I have but makes the `noexec` optional. i.e. our task will never downgrade the system security by removing nosuid or nodev even if you disable those tasks and the tasks runs if any one of the vars is enabled.

```yaml
- name: |
      "LOW | RHEL-07-021022 | PATCH | The Red Hat Enterprise Linux operating system must mount /dev/shm with the nodev option."
      "LOW | RHEL-07-021023 | PATCH | The Red Hat Enterprise Linux operating system must mount /dev/shm with the nosuid option."
      "LOW | RHEL-07-021024 | PATCH | The Red Hat Enterprise Linux operating system must mount /dev/shm with the noexec option."
  mount:
      path: /dev/shm
      state: mounted
      src: "{{ dev_shm_mount.device|default('tmpfs') }}"
      fstype: "{{ dev_shm_mount.fstype|default('tmpfs') }}"
      opts: "{{ dev_shm_mount_opts }}"
  when: (rhel_07_021022 or rhel_07_021023 or rhel_07_021024)
  vars:
      dev_shm_mount: "{{ ansible_mounts | json_query('[?mount == `/dev/shm`] | [0]') }}"
      dev_shm_mount_opts: "{{ dev_shm_mount.options|default('defaults') }},nodev,nosuid{{ (rhel_07_021024) | ternary(',noexec', '') }}"
  tags:
      - RHEL-07-021022
      - RHEL-07-021023
      - RHEL-07-021024
```

jamescassell:
I like your proposed alternative.  Thanks for the explanation.

shepdelacreme:
> I like your proposed alternative. Thanks for the explanation.
> 
> Edit: does 'defaults' resolve to 'rw,seclabel'? Othewise, maybe we should "default" to that rather than 'defaults'.

Yup just tested and defaults does indeed resolve to 'rw,seclabel'

tasks/fix-cat3.yml
```diff
@@ -32,6 +32,25 @@
...
+      "LOW | RHEL-07-021022 | PATCH | The Red Hat Enterprise Linux operating system must mount /dev/shm with the nodev option."
+      "LOW | RHEL-07-021023 | PATCH | The Red Hat Enterprise Linux operating system must mount /dev/shm with the nosuid option."
+      "LOW | RHEL-07-021024 | PATCH | The Red Hat Enterprise Linux operating system must mount /dev/shm with the noexec option."
+  mount:
+      path: /dev/shm
+      state: mounted
+      src: "{{ dev_shm_mount.device|default('tmpfs') }}"
+      fstype: "{{ dev_shm_mount.fstype|default('tmpfs') }}"
+      opts: "{{ dev_shm_mount_opts }}"
+  when: (rhel_07_021022 or rhel_07_021023 or rhel_07_021024)
```
defionscode: style nitpick, I'm not fond of using parentheses in `when` unless your are splitting out logic for conditional evaluation
```suggestion
  when: rhel_07_021022 or rhel_07_021023 or rhel_07_021024
```

tasks/fix-cat3.yml
```diff
@@ -32,6 +32,25 @@
...
+  vars:
+      dev_shm_mount: "{{ ansible_mounts | json_query('[?mount == `/dev/shm`] | [0]') }}"
+      dev_shm_mount_opts: "{{ dev_shm_mount.options|default('defaults') }},nodev,nosuid{{ (rhel_07_021024) | ternary(',noexec', '') }}"
```
defionscode: Style consistency for j2 
```suggestion
      dev_shm_mount_opts: "{{ dev_shm_mount.options | default('defaults') }},nodev,nosuid{{ (rhel_07_021024) | ternary(',noexec', '') }}"
```

PR #184, opened 2018-10-19
Fixes deprecation warnings that are associated with loops using yum

borener:
Description of PR
Fixes Deprecation warnings with the upgrade to Ansible 2.7.
Specifically it removes the loop that is no longer needed with the upgrade.

Type of change
Feature Pull Request

Fixes an issue

Fixes following Deprecation Warning

[DEPRECATION WARNING]: Invoking "yum" only once while using a loop via squash_actions is deprecated. Instead of using a loop to supply multiple items and specifying name: {{ item }}, please use name: [u'snappy', u'snappy- devel', u'zlib', u'zlib-devel', u'hwloc', u'gcc', u'lzo', u'lzo-devel', u'python', u'make'] and remove the loop. This feature will be removed in version 2.11. Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.

### DIFF
--- a/tasks/prelim.yml
+++ b/tasks/prelim.yml
@@ -200,11 +200,10 @@
 
 - name: "PRELIM | RHEL-07-020210 RHEL-07-020220| Install SELinux related dependencies"
   yum:
-      name: "{{ item }}"
-  with_items:
-      - libselinux-python
-      - selinux-policy-targeted
-      - selinux-policy-mls
+      name:
+        - libselinux-python
+        - selinux-policy-targeted
+        - selinux-policy-mls
   when:
       - rhel_07_020210 or rhel_07_020220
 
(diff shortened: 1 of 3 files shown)


tasks/prelim.yml
```diff
@@ -200,11 +200,10 @@
 
 - name: "PRELIM | RHEL-07-020210 RHEL-07-020220| Install SELinux related dependencies"
   yum:
-      name: "{{ item }}"
-  with_items:
-      - libselinux-python
-      - selinux-policy-targeted
-      - selinux-policy-mls
+      name:
+        - libselinux-python
```
jamescassell: Needs 2 more spaces of indent, per our guidelines.

PR #190, opened 2018-10-27
fix RHEL-07-010010

### LINKED ISSUE #189: RHEL-07-010010 failing oscap scan due to quoting problems in task
The following task returns nothing to `stdout` and thus fails to trigger updates on the subsequent command.

```yaml
      - name: "HIGH | RHEL-07-010010 | AUDIT | The file permissions, ownership, and group membership of system files and commands must match the vendor values."
        shell: 'rpm -Va --nolinkto --nofiledigest --nosize --nomtime --nodigest --nosignature | grep -E ''^(.M|.....U|......G)'' | tee /dev/stderr | cut -c13- | sed ''s/^ //'' | xargs rpm -qf --qf=''%{name}\n'' | sort -u'
        args:
            warn: no
        check_mode: no
        failed_when: no
        changed_when: rhel_07_010010_audit.stdout != ""
        register: rhel_07_010010_audit
```

I have a fix ready, PR inbound

defionscode:
fixes #189

### DIFF
--- a/tasks/fix-cat1.yml
+++ b/tasks/fix-cat1.yml
@@ -1,7 +1,8 @@
 ---
 - block:
       - name: "HIGH | RHEL-07-010010 | AUDIT | The file permissions, ownership, and group membership of system files and commands must match the vendor values."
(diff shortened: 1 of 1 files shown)


defionscode:
My bet is that it's more of an issue with Ansible's parsing than anything else, and there is no new line added, with the current version of Ansible you'd need to do `|+` to do the implicit newline at the end of the text. I didnt do much digging outside of just validation with `oscap eval ...`

PR #191, opened 2018-10-29
Some light updates

defionscode:
Summary

- Fixes 010290 which was failing SCAP checks due to password-auth not being remediate for the presence of nullok

- Add a capability to pass 030300

- Fixed a dup auditd for 030740 and made it so that rule would pass. It _was_ failing due to the presence of `perm-x` which is not part of the official guidance.

### DIFF
--- a/tasks/audit_command.yml
+++ b/tasks/audit_command.yml
@@ -12,6 +12,6 @@
       state: "{{ audit_present | ternary('present', 'absent') }}"
   vars:
       trivial_audit: "-w {{ item.path }} -p x {{ rhel7stig_workaround_for_ssg_benchmark | ternary('', '-F auid!=4294967295 ') }}-k {{ item.key }}"
-      normal_audit: "-a always,exit -F path={{ item.path }} -F perm=x -F auid>=1000 -F auid!=4294967295 -k {{ item.key }}"
+      normal_audit: "-a always,exit -F path={{ item.path }} {% if not item.no_perm_x_filter is defined or not item.no_perm_x_filter %}-F perm=x{% endif %} -F auid>=1000 -F auid!=4294967295 -k {{ item.key }}"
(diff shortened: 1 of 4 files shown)


tasks/fix-cat2.yml
```diff
@@ -1301,9 +1301,9 @@
 
 - name: "MEDIUM | RHEL-07-030300 | PATCH | The operating system must off-load audit records onto a different system or media from the system being audited."
   lineinfile:
-    path: /etc/audisp/audisp-remote.conf
-    regexp: ^remote_server +=
-    line: remote_server = {{ rhel7stig_audisp_remote_server }}
+      path: /etc/audisp/audisp-remote.conf
+      regexp: ^remote_server +=
```
```suggestion
      regexp: ^remote_server *=
```

tasks/audit_command.yml
```diff
@@ -12,6 +12,6 @@
       state: "{{ audit_present | ternary('present', 'absent') }}"
   vars:
       trivial_audit: "-w {{ item.path }} -p x {{ rhel7stig_workaround_for_ssg_benchmark | ternary('', '-F auid!=4294967295 ') }}-k {{ item.key }}"
-      normal_audit: "-a always,exit -F path={{ item.path }} -F perm=x -F auid>=1000 -F auid!=4294967295 -k {{ item.key }}"
+      normal_audit: "-a always,exit -F path={{ item.path }} {% if not item.no_perm_x_filter is defined or not item.no_perm_x_filter %}-F perm=x{% endif %} -F auid>=1000 -F auid!=4294967295 -k {{ item.key }}"
```
jamescassell:
```suggestion
      normal_audit: "-a always,exit -F path={{ item.path }} {% if not item.no_perm_x_filter is defined or not item.no_perm_x_filter %}-F perm=x {% endif %}-F auid>=1000 -F auid!=4294967295 -k {{ item.key }}"
```

PR #200, opened 2018-12-19
Fix up naming, block statements, and make 010260 use disruption-high mode for doc gen

shepdelacreme:
These fixes/changes are mainly to support the documentation generation that is in the `migrate_ah_docs` branch 725b2bf.

The changes split by atomic commits so it is easy to review. There are name updates to bring them in line with the standard, addition of a name parameter so there are no bare `block` statements, and the last change makes the RHEL-07-010260 task be a disruption-high item instead of just being disabled by default since the reasoning for disabling it was that it has a high chance of locking out users.

Once this is merged I can work on getting the doc generation merged in and running for the devel branch.

### DIFF
--- a/tasks/fix-cat1.yml
+++ b/tasks/fix-cat1.yml
@@ -432,7 +432,8 @@
       - RHEL-07-040540
       - shosts
 
-- block:
+- name: "HIGH | RHEL-07-040550 | PATCH | There must be no shosts.equiv files on the system."
+  block:
       - name: "HIGH | RHEL-07-040550 | AUDIT | There must be no shosts.equiv files on the system."
         command: find {{ rhel7stig_local_mounts | join(' ') }} -xdev -name 'shosts.equiv'
         check_mode: no
--- a/tasks/fix-cat1.yml
+++ b/tasks/fix-cat1.yml
@@ -1,5 +1,5 @@
 ---
-- name: "HIGH | RHEL-07-010010 | PATCH | The file permissions, ownership, and group membership of system files and commands must match the vendor values."
+- name: "HIGH | RHEL-07-010010 | The file permissions, ownership, and group membership of system files and commands must match the vendor values."
   block:
       - name: "HIGH | RHEL-07-010010 | AUDIT | The file permissions, ownership, and group membership of system files and commands must match the vendor values."
         shell: |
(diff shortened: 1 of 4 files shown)


tasks/fix-cat1.yml
```diff
@@ -1,5 +1,6 @@
 ---
-- block:
+- name: "HIGH | RHEL-07-010010 | PATCH | The file permissions, ownership, and group membership of system files and commands must match the vendor values."
```
jamescassell: I wouldn't put "PATCH" as part of the block title, but probably not worth changing it for this PR; only going forward.

tasks/fix-cat1.yml
```diff
@@ -124,7 +124,6 @@
 
 
 - name: |
-
```
jamescassell: I understand the reason this newline was added, but I'm torn as to whether to include it.  W/o it, the two following lines don't line up on execution, but it looks messy in the playbook...  FWIW, I've been not including this newline in task blocks I've written, consistent w/ your change here.

tasks/fix-cat2.yml
```diff
@@ -1032,7 +1032,7 @@
   tags:
       - RHEL-07-020690
 
-- name: "MEDIUM | RHEL-07-020700 | Local initialization files for local interactive users must be group-owned by the users primary group or root."
+- name: "MEDIUM | RHEL-07-020700 | PATCH | Local initialization files for local interactive users must be group-owned by the users primary group or root."
```
jamescassell: I'd probably skip adding these... benchparse also doesn't include them at the block level, IIRC.

tasks/fix-cat2.yml
```diff
@@ -317,14 +317,19 @@
...
         command: chage -M 60 {{ item }}
+        check_mode: "{{ rhel7stig_disruptive_check_mode }}"
         with_items: "{{ rhel_07_010260_audit.stdout_lines }}"
-  when: rhel_07_010260
+  when:
+      - rhel_07_010260
+      - rhel7stig_disruptive
   tags:
       - RHEL-07-010260
+      - disruption-high
```
jamescassell: Awesome, thanks!

tasks/fix-cat2.yml
shepdelacreme: hmm ok good to know. I did scrub through and change all the block titles to include PATCH but I can back those changes out easily if that isn't the desired end state. I thought I might need the name titles to be more consistent for doing parsing in the doc generator but really I just need the | RHEL-07-###### | identifier to be there in the second spot.

So if the consensus is no PATCH identifier in block titles then I will back those changes out of this PR.

tasks/fix-cat1.yml
shepdelacreme: Yes I think I remember this discussion now. I removed it because it made it easier for the doc generator to parse. I'm one to lean towards not including it since I don't typically watch task output in the console.

shepdelacreme:
Removed the changes to existing `block` statement names and standardized the rest on `SEV | ID | Title` format.

PR #201, opened 2018-12-19
Merge in documentation generation stuff

shepdelacreme:
This is a placeholder PR for merging in the documentation generation so I can test CI udpates and what not. #200 needs to be merged in to devel first and this PR should be rebased and merged in to avoid the addition of all this cruft commits here.


shepdelacreme:
Removed the WIP tag as this is ready to merge in once reviewed and after #200 gets merged.

shepdelacreme:
For the checksums...do you mean running the compare during the `make` build process? I'm not sure I see the benefit since all changes to either file would be tracked in source control but I could add that easily for the Apache file. The XCCDF file I'd like to eventually not store in the repo at all and use `stig-cli` or pull, unzip, and parse directly from the DISA site each time.

shepdelacreme:
I see. Ran a diff on the Apache license and fixed the only differences which were minor. i.e.  {} -> [] from the official version. DISA doesn't publish the checksums for the XCCDF files that I can find but I checked the local version against a fresh DISA download and they matched.

`sha256 -> ec66a1c2c70db9885f4fa07cc7c28ed7a9f5dd0a203f4bca7b47e1a6f8489c2e`

Also by updating that license I realize I just reset the review requirements for the PR.

jamescassell:
This might be a good candidate for squash/merge, for reasons you mentioned in the original comment.

PR #202, opened 2018-12-21
update travis to pull from new RH registry loc with auth

shepdelacreme:
Red Hat is migrating their container registry from `registry.access.redhat.com` to `registry.redhat.io` and going forward will require auth even for pull operations. This updates the Travis config to account for that.


shepdelacreme:
It is a bit early but I have seen some intermittent CI failures on pull from the old registry location. When I saw this notice from RH I decided to switch early to see if it clears it up.

PR #208, opened 2019-02-13
SSH key mode

jamescassell:
Partial fix for V2R1

### DIFF
--- a/tasks/fix-cat2.yml
+++ b/tasks/fix-cat2.yml
@@ -1971,9 +1971,9 @@
       - RHEL-07-040410
       - ssh
 
-- name: "MEDIUM | RHEL-07-040420 | The SSH private host key files must have mode 0600 or less permissive."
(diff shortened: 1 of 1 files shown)


jamescassell:
Any idea why the CI failed?

shepdelacreme:
Hmm looks like just the Docker tests that failed. I’ll dig into it this weekend.

shepdelacreme:
LGTM...I'll review and approve once the conflicts are resolved.

jamescassell:
Indeed, I didn't anticipate that merging the large one would conflict all the rest, since there is no logical conflict... I'll fix it up soon.

shepdelacreme:
RHSM shouldn't fail anymore. I worked around the issue by creating a new user that could actually accept the RH terms.

jamescassell:
conflict is period vs no period...

jamescassell:
Looks like all tests are passing now after your workaround.  (If you're feeling ambitious, you could open a RH support case for the issue?)  Also there are no remaining conflicts on my (6) PRs so they should be merge-able.

Issue #174, opened 2018-09-26
Update for v2r1


jamescassell:
Here's a fairly comprehensive list:

### Global Changes

- [x] Every rule title was changed to include "Red Hat Enterprise Linux" in some form
- [x] Throughout, a required, but commented-out configuration setting fails (this is obvious; will handle compliance on a rule-by-rule basis; some rule are okay w/ unconfigured setting)

### Modified Rules

- [ ] RHEL-07-010010: if non-RPM-verified (user/group ownership, permissions) files are not documented, it is a finding (should subtract a list of documented bad files and/or packages from the list of found bad files and/or packages)
- [ ] RHEL-07-010250: "If passwords are not being used for authentication, this is Not Applicable."
- [ ] RHEL-07-010310: "If passwords are not being used for authentication, this is Not Applicable."
- [ ] RHEL-07-010500: "--smartcardaction=1" -> "--smartcardaction=0"
- [ ] RHEL-07-020020: "If an HBSS or HIPS is active on the system, this is Not Applicable."
- [ ] RHEL-07-020210: "If an HBSS or HIPS is active on the system, this is Not Applicable."
- [ ] RHEL-07-020220: "If an HBSS or HIPS is active on the system, this is Not Applicable."
- [ ] RHEL-07-020230: "ctrl.alt.del.target is masked and not active" -- "not active" is new
- [ ] RHEL-07-020730: use '-xdev' when finding world-writable files (rule not yet implemented)
- [ ] RHEL-07-021020: nosuid on "file systems that are being imported via Network File System (NFS)." - mostly a wording change
- [ ] RHEL-07-031000: "/etc/rsyslog.conf" -> '"/etc/rsyslog.conf" or an "/etc/rsyslog.d/*.conf"' (should move to config snippet, add tombstone for content in rsyslog.conf main)
- [ ] RHEL-07-032000: install anti-virus -- removed reference to McAfee
- [ ] RHEL-07-040180: "If LDAP is not being utilized, this requirement is Not Applicable.", "ssl = start_tls" -> "ldap_id_use_start_tls = true", /etc/pam_ldap.conf -> /etc/sssd/sssd.conf (should create implicit_files domain explicitly to enable sssd generically and configure)
[...]

PR #211, opened 2019-02-14
audit suid/sgid executions - Fix RHEL-07-030360 for V2R1 and V2R2

JSurf:
Changes to be compliant with changed RHEL-07-030360 in V2R1 and V2R2

**New Text since V2R1:**
Configure the operating system to audit the execution of privileged functions.

Add or update the following rule in "/etc/audit/rules.d/audit.rules":

Note: The rules are duplicated to cover both 32-bit and 64-bit architectures. Only the lines appropriate for the system architecture must be configured.

-a always,exit -F arch=b32 -S execve -C uid!=euid -F euid=0 -k setuid
-a always,exit -F arch=b64 -S execve -C uid!=euid -F euid=0 -k setuid
-a always,exit -F arch=b32 -S execve -C gid!=egid -F egid=0 -k setgid
-a always,exit -F arch=b64 -S execve -C gid!=egid -F egid=0 -k setgid

The audit daemon must be restarted for the changes to take effect.

**Old Text: in V1R4**
Configure the operating system to audit the execution of privileged functions.

To find the relevant "setuid"/"setgid" programs, run the following command for each local partition [PART]:

find [PART] -xdev -type f \( -perm -4000 -o -perm -2000 \) 2>/dev/null

For each "setuid"/"setgid" program on the system, which is not covered by an audit rule for a (sub) directory (such as "/usr/sbin"), add a line of the following form to "/etc/audit/rules.d/audit.rules", where <suid_prog_with_full_path> is the full path to each "setuid"/"setgid" program in the list:

-a always,exit -F path=<suid_prog_with_full_path> -F perm=x -F auid>=1000 -F auid!=4294967295 -k setuid/setgid

### DIFF
--- a/tasks/fix-cat2.yml
+++ b/tasks/fix-cat2.yml
@@ -1382,22 +1382,27 @@
 
 - name: "MEDIUM | RHEL-07-030360 | The Red Hat Enterprise Linux operating system must audit all executions of privileged functions."
   block:
-      - name: "MEDIUM | RHEL-07-030360 | AUDIT | The Red Hat Enterprise Linux operating system must audit all executions of privileged functions. (find suid/sgid programs)"
-        command: find {{ rhel7stig_local_mounts | join(' ') }} -xdev -type f ( -perm -4000 -o -perm -2000 )
-        check_mode: no
-        changed_when: no
-        register: rhel_07_030360_audit
-
-      - name: "MEDIUM | RHEL-07-030360 | The Red Hat Enterprise Linux operating system must audit all executions of privileged functions."
-        lineinfile:
+      - name: "MEDIUM | RHEL-07-030360 | PATCH | The Red Hat Enterprise Linux operating system must audit all executions of privileged functions. (Delete obsolete suid/sgid rules)"
+        file:
             path: "/etc/audit/rules.d/rhel7stig_suid_sgid_commands.rules"
+            state: absent
+
+      - name: "MEDIUM | RHEL-07-030360 | PATCH | The Red Hat Enterprise Linux operating system must audit all executions of privileged functions. (Audit rules for execve calls)"
+        lineinfile:
+            path: "/etc/audit/rules.d/rhel7stig_execve_calls.rules"
             create: yes
             owner: root
             group: root
             mode: 0600
-            line: "-a always,exit -F path={{ item }} -F perm=x -F auid>=1000 -F auid!=4294967295 -k setuid/setgid"
+            line: "{{ item }}"
             state: present
-        with_items: "{{ rhel_07_030360_audit.stdout_lines }}"
+        with_items:
+            - "-a always,exit -F arch=b32 -S execve -C uid!=euid -F euid=0 -k setuid"
+            - "-a always,exit -F arch=b64 -S execve -C uid!=euid -F euid=0 -k setuid"
+            - "-a always,exit -F arch=b32 -S execve -C gid!=egid -F egid=0 -k setgid"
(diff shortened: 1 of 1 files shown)


JSurf:
Fixes RHEL-07-030360 from #174 Update for v2r1

tasks/fix-cat2.yml
```diff
@@ -1380,24 +1380,29 @@
...
-        check_mode: no
-        changed_when: no
-        register: rhel_07_030360_audit
+      - name: "Delete obsolete suid/sgid rules"
+        file:
+            path: "/etc/audit/rules.d/rhel7stig_suid_sgid_commands.rules"
+            state: absent
 
-      - name: "MEDIUM | RHEL-07-030360 | All privileged function executions must be audited."
+      - name: "Audit rules for execve calls"
```
defionscode: I think we want to keep the name the same it helps with mapping output to specific rules

tasks/fix-cat2.yml
```diff
@@ -1380,24 +1380,29 @@
...
+            path: "/etc/audit/rules.d/rhel7stig_execve_calls.rules"
             create: yes
             owner: root
             group: root
             mode: 0600
-            line: "-a always,exit -F path={{ item }} -F perm=x -F auid>=1000 -F auid!=4294967295 -k setuid/setgid"
+            line: "{{ item }}"
             state: present
-        with_items: "{{ rhel_07_030360_audit.stdout_lines }}"
+        with_items:
```
defionscode: I wonder if maybe we should start introducing the usage of `loop` instead of `with_items`. It's not terribly important at the moment, maybe it can just be part of a larger modernization refactor to keep in line with changes in Ansible upstream

jamescassell: Would break ansible 2.4. with_<lookup> is not deprecated yet.

defionscode: Ansible 2.4 is EOL https://docs.ansible.com/ansible/latest/reference_appendices/release_and_maintenance.html#release-status

tasks/fix-cat2.yml
JSurf: Changed names of block tasks, see latest commit, what do you think about this ? Not sure if you like the extra comments in the block names (Delete obsolete suid/sgid rules) and (Audit rules for execve calls)

tasks/fix-cat2.yml
```diff
@@ -1380,24 +1380,29 @@
...
+            - "-a always,exit -F arch=b32 -S execve -C uid!=euid -F euid=0 -k setuid"
+            - "-a always,exit -F arch=b64 -S execve -C uid!=euid -F euid=0 -k setuid"
+            - "-a always,exit -F arch=b32 -S execve -C gid!=egid -F egid=0 -k setgid"
+            - "-a always,exit -F arch=b64 -S execve -C gid!=egid -F egid=0 -k setgid"
```
shepdelacreme: These lines are duplicated. i.e. both `arch=b32` and `arch=b64` will be inserted regardless of the underlying system arch. Per the STIG item, I think there should only be one or the other? i.e. `Only the lines appropriate for the system architecture must be present`.

You should be able to do this using hostvars from the system.

jamescassell:
@JSurf can you rebase against devel, or so you mind if I do?

JSurf:
Done rebasing and squashed to one commit

tasks/fix-cat2.yml
jamescassell: b32 and b64 are both appropriate for x86_64 because it has both versions of the syscall.

shepdelacreme: The DISA content is not clear on this point so thanks!

PR #214, opened 2019-02-19
temp sshd keys for config validation

jamescassell:
if the keys don't exist, "sshd -t -f %s" fails

remove the keys at the end to allow the normal key generation process to take place

### DIFF
--- a/handlers/main.yml
+++ b/handlers/main.yml
@@ -50,6 +50,14 @@
   command: auditctl -f {{ rhel7stig_auditd_failure_flag }}
   failed_when: no
 
+- name: clean up ssh host key
+  file:
+      path: "{{ item }}"
+      state: absent
+  with_items:
+      - /etc/ssh/ssh_host_rsa_key
+      - /etc/ssh/ssh_host_rsa_key.pub
+
 - name: init aide and wait
   command: /usr/sbin/aide --init -B 'database_out=file:{{ rhel7stig_aide_temp_db_file }}'
   notify: move aide db
--- a/tasks/prelim.yml
+++ b/tasks/prelim.yml
@@ -213,6 +213,12 @@
         yum:
             name: openssh-server
 
+      - name: PRELIM | Start SSH
+        service:
+            name: sshd
+            state: "{{ rhel7stig_service_started }}"
+            enabled: yes
+
       - name: PRELIM | check if ssh host key exists
         stat:
             path: /etc/ssh/ssh_host_rsa_key
--- a/tasks/prelim.yml
+++ b/tasks/prelim.yml
@@ -207,7 +207,19 @@
   when:
       - rhel_07_020210 or rhel_07_020220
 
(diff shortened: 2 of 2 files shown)


handlers/main.yml
```diff
@@ -50,6 +50,14 @@
...
   failed_when: no
 
+- name: clean up ssh host key
+  file:
+      path: "{{ item }}"
+      state: absent
+  with_items:
+      - /etc/ssh/ssh_host_rsa_key
+      - /etc/ssh/ssh_host_rsa_key.pub
+
```
shepdelacreme: What are the implications of cleaning up this key? If I expect this role to install, enable, and configure OpenSSH Server then removing these keys means I will need to regenerate them again? Or will they get generated the next time ssh starts up?

jamescassell: They are generated at boot time. I'm running the role at kickstart time, before they've been generated. I clean them up so that machines deployed in "gold image" style will still generate fresh keys at first boot. The role only cleans these keys if it created them.

shepdelacreme: Right...some people might not be using the role in the same way and may expect that the keys exist (even if the role creates them). Typically in a kickstart setup, I would have all my cleanup items like this at the end, outside of this role. Maybe it should just be optional or something. `rhel7stig_cleanup_keys` or `rhel7stig_cleanup_for_gold` or something. The second option can be used if we have any other cleanup items that should be run before "gold".

jamescassell: If we create the keys, we should clean them up, i'm not opinion. I don't feel comfortable that my invocation is fips compliant, and I'd rather let the system do it. I also create only one type of key, whereas the system generates 3.

I'd be okay gating this behind a flag, disabled by default, in which case the role fails on any ssh remediation.

shepdelacreme: I thought OpenSSH would generate the host keys when it starts up if they don't exist. Can't we just start the service so it generates the keys? Then you can have a conditional clean up task and others that want to keep the keys can keep them?

jamescassell: I'd be fine inserting 'service: name=sshd state=started' before the manual key creation, so that it will  auto-create the keys in any case except for running in a kickstart environment.  (starting services in the ks environment does not work, and we already have a flag to avoid trying. `system_is_chroot`)  The latest version of ansible detects this automatically and issue a warning if you try.

shepdelacreme: I've changed my mind on this. On further consideration, it is very unlikely that someone is running this role against a system that does not have SSH installed unless they are doing it as part of kickstart or some other golden image creation process that will get a reboot after.

shepdelacreme:
@jamescassell I'm fine without the early service start actually. Sorry. I reconsidered the situation and the way you were doing it before is fine.

jamescassell:
Starting SSH already happens in RHEL-07-040310.  It's probably cleaner to have it done here ahead of time and also avoids having to manually create SSH keys in more cases.

PR #215, opened 2019-02-19
gather necessary facts on demand

jamescassell:
remove requirement to run role with 'gather_facts: yes'

### DIFF
--- a/tasks/fix-cat2.yml
+++ b/tasks/fix-cat2.yml
@@ -1093,6 +1093,19 @@
       - RHEL-07-020730
       - notimplemented
 
+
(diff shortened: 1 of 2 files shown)


shepdelacreme:
My concern with this is that we now need to add tasks to gather facts each time we need them. Does doing lazy fact gathering gain you much in terms of playbook execution time?

jamescassell:
I'd be fine with this being "not supported" but we'll take patches if they don't break anyone else... on some hardware systems, fact gathering is quite slow.

shepdelacreme:
Ok...I'm fine with this if it saves a good amount of time during playbook run. I just never thought of fact gathering as the slow part of Ansible roles. If we go this route I think we should standardize on it.

jamescassell:
We agreed during the meeting today that we will support 'gather_facts: no' and that we will change testing to account/test for this.

PR #220, opened 2019-02-25
pam_lastlog: "silent" most not be present

jamescassell:
If "pam_lastlog" is missing from "/etc/pam.d/postlogin" file, or the
silent option is present, this is a finding.

A compliant line is already placed by the role, but the DISA benchmark
is flagging another default line as non-compliant.

I'd appreciate any testing against existing setups; things worked fine for me w/ this change.

### DIFF
--- a/tasks/fix-cat3.yml
+++ b/tasks/fix-cat3.yml
@@ -143,6 +143,15 @@
 
 - name: "LOW | RHEL-07-040530 | The Red Hat Enterprise Linux operating system must display the date and time of the last successful account logon upon logon."
   block:
+      - name: "LOW | RHEL-07-040530 | Update pam_lastlog control to satisfy benchmark."
+        pamd:
+            name: postlogin
+            type: session
+            control: "{{ old_control }}"
+            new_control: "{{ default_control }}"
+            module_path: pam_lastlog.so
+            state: updated
+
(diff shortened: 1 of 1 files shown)


jamescassell:
It generally feels cleaner to use the pamd module, but that would require some testing to make sure it works with common configurations.

jamescassell:
Closing for now as I think there's a better way:
- The postlogin config has been identical since RHEL 7.0
- We should instead modify this config to be compliant and clear out our custom blockinfile

Here's the postlogin script in all version of RHEL 7:
```
#%PAM-1.0
# This file is auto-generated.
# User changes will be destroyed the next time authconfig is run.


session     [success=1 default=ignore] pam_succeed_if.so service !~ gdm* service !~ su* quiet
session     [default=1]   pam_lastlog.so nowtmp showfailed
session     optional      pam_lastlog.so silent noupdate showfailed
```

jamescassell:
I've force-pushed a better solution.  This mutates the default postlogin config to pass the DISA benchmark.  Prior to V2R1, it was permissible to have the 'silent' option enabled as long as SSH itself printed the lastlog info.  In that update, this allowance was removed: https://vaulted.io/library/disa-stigs-srgs/red_hat_enterprise_linux_7_security_technical_implementation_guide/V-72275?version=V1R4&compareto=V2R2

(By the way, I came across that really helpful website that would have save lots of time when V2R1 came out if I'd known about it then.)

shepdelacreme:
Wow! That IS a handy website! Thanks for the link.

PR #235, opened 2019-04-04
fix local.yml for ansible-2.4

jamescassell:
ansible_python_interpreter is already set automatically by ansible
exactly as this line attempts to do

### DIFF
--- a/local.yml
+++ b/local.yml
@@ -3,7 +3,6 @@
   connection: local
   become: yes
   vars:
-      ansible_python_interpreter: "{{ ansible_playbook_python }}"
       is_container: false
   roles:
       - role: "{{ playbook_dir }}"


shepdelacreme:
I'm pretty sure it is related to this issue: http://willthames.github.io/2018/07/01/connection-local-vs-delegate_to-localhost.html

jamescassell:
when host is the implicit `localhost`, ansible automatically sets this variable, as was done explicitly here.  It's a poorly-documented bit of magic, but it works at least back to 2.5.

shepdelacreme:
I'm going to test this locally with Vagrant to make sure it all still works as expected. I want to say this is a holdover from pre-2.4 times but the post I linked above isn't that old and I know I've run across this issue myself, just not in the last 6 months :)

jamescassell:
Thanks for verifying!

PR #236, opened 2019-04-19
update supported OS versions

jamescassell:


defaults/main.yml
```diff
@@ -291,8 +291,8 @@ rhel7stig_gui: no
 # These are the minimum supported releases.
 # (Red Hat has support for older versions if you pay extra for it.)
 rhel7stig_min_supported_os_ver:
-    RedHat: "7.5"
-    CentOS: "7.4.1708"
+    RedHat: "7.6"
```
dirtyharrycallahan: Wouldn't it be appropriate to detect if the platform was a current EUS version? 7.4 EUS will be supported until August 2019.

jamescassell: PRs welcome. You can also override these vars on a per-host or per-group basis in your ansible inventory.

PR #245, opened 2019-06-21
Implement GUI items

dfederlein:
Pull Request 2: Electric Boogaloo.

Added some tasks for our testing process to evaluate, fixed linter issues (hopefully.)

@shepdelacreme let me know if you want to chat about these today.


dfederlein:
> Is there a good reason not to use the `dconf` module?


My first response was to template the files/vars into the configs as instructed by STIG's remediation then dconf update. When I read the docs on that module a simple dconf update wasn't in the examples (that I saw.)   If there's a cleaner way to do it I'm open.

jamescassell:
> > Is there a good reason not to use the `dconf` module?
> 
> My first response was to template the files/vars into the configs as instructed by STIG's remediation then dconf update. When I read the docs on that module a simple dconf update wasn't in the examples (that I saw.) If there's a cleaner way to do it I'm open.

In general, the STIG fix text is just a guideline of "here's one way maybe to fix it"... I don't have time this moment, but I checked yesterday and it looked like the module should suffice.

shepdelacreme:
I'm good with using `template` here instead of the `dconf` module. It is a file that contains no other settings other than the two that need to be set by the STIG so it is not likely to stomp on existing configs and it is much simpler/cleaner than dealing with the `dconf` module which is a bit of mess.

tasks/fix-cat2.yml
```diff
@@ -1,25 +1,27 @@
...
+        register: rhel_07_010040dconfout
+        changed_when: rhel_07_010040dconfout.rc <= 1
+        failed_when: rhel_07_010040dconfout.rc > 1
   when:
       - rhel7stig_gui
+      - rhel_07_010030
       - rhel_07_010040
   tags:
-      - RHEL-07-010040
+      - RHEL-07-010030
```
shepdelacreme: This should have both STIG IDs as tags. i.e. add `RHEL-07-010040`

dfederlein: committed.

dfederlein:
> > > Is there a good reason not to use the `dconf` module?
> > 
> > 
> > My first response was to template the files/vars into the configs as instructed by STIG's remediation then dconf update. When I read the docs on that module a simple dconf update wasn't in the examples (that I saw.) If there's a cleaner way to do it I'm open.
> 
> In general, the STIG fix text is just a guideline of "here's one way maybe to fix it"... I don't have time this moment, but I checked yesterday and it looked like the module should suffice.

This is meant to be easy for non-ansible to follow and audit though. Auditors don't get creative with interpretations. My intention was output actions in the playbook run they can literally check off as it goes. Sometimes not being the most elegant/direct is good in support of cross-functional teams understanding what's happening.

On to the dconf module: like the pamd module it has its issues, but the big thing is that it's a community maintained module and when possible I favor core maintained modules when I can do something for supportability. I don't want to be the one fixing the dconf module on upgrade to new ansible behaviors, or waiting on community to fix it. I can engage a direct remedy with Red Hat support when I have an issue.

tasks/fix-cat2.yml
```diff
@@ -1,25 +1,28 @@
...
-      - rhel_07_010030
-  tags:
-      - RHEL-07-010030
-      - dod_logon_banner
-      - notimplemented
+### RHEL-07-010030 | RHEL-07-010040 combined as related tasks in regards to a config file no other content will be in.
+- name: "MEDIUM | RHEL-07-010030 | RHEL-07-010040 | PATCH | The Red Hat Enterprise Linux operating system must display the Standard Mandatory DoD Notice and Consent Banner before granting local or remote access to the system via a graphical user logon."
+  block:
+      - name: "MEDIUM | RHEL-07-010030 | RHEL-07-010040 | PATCH | The Red Hat Enterprise Linux operating system must display the Standard Mandatory DoD Notice and Consent Banner before granting local or remote access to the system via a graphical user logon."
+        template:
```
jamescassell: I'd tend to prefer `copy: content` since the template is trivial.

tasks/fix-cat2.yml
```diff
@@ -40,54 +43,107 @@
...
 - name: "MEDIUM | RHEL-07-010070 | PATCH | The Red Hat Enterprise Linux operating system must initiate a screensaver after a 15-minute period of inactivity for graphical user interfaces."
-  command: "true"
-  changed_when: no
+  block:
+      - name: "MEDIUM | RHEL-07-010070 | PATCH | The Red Hat Enterprise Linux operating system must initiate a screensaver after a 15-minute period of inactivity for graphical user interfaces."
+        lineinfile:
+            create: yes
+            dest: /etc/dconf/db/local.d/00-screensaver
+            insertafter: '^#[org/gnome/desktop/screensaver]'
+            regexp: '^#?idle-delay'
```
jamescassell: I generally tend to prefer an `insertafter` the commented version and a `regexp` for the non-commented version, but not a huge deal.

tasks/fix-cat2.yml
```diff
@@ -2376,7 +2466,7 @@
         when:
             - rhel_07_040680_rpm_audit.rc == 0
             - rhel_07_040680_postconf_audit.stdout != 'smtpd_client_restrictions = permit_mynetworks, reject'
-  when: rhel_07_040680
+
```
jamescassell: What was the reason for this change? (This is the only one I'd like answered for sure before merging)

jamescassell:
> I'm good with using `template` here instead of the `dconf` module. It is a file that contains no other settings other than the two that need to be set by the STIG so it is not likely to stomp on existing configs and it is much simpler/cleaner than dealing with the `dconf` module which is a bit of mess.

Ack.  I guess the issue is that `dconf` module doesn't provide a way to say which db to modify?

dfederlein:
> > I'm good with using `template` here instead of the `dconf` module. It is a file that contains no other settings other than the two that need to be set by the STIG so it is not likely to stomp on existing configs and it is much simpler/cleaner than dealing with the `dconf` module which is a bit of mess.
> 
> Ack. I guess the issue is that `dconf` module doesn't provide a way to say which db to modify?

That's one issue, yes.  I did realize about the same time you were writing the comments on the dconf command that I needed to conditionalize it.  while I didn't use a handler (and I'm open to talking about whether a handler is more appropriate) I did put in a conditional based on the changed status of the previous task.

tasks/fix-cat2.yml
dfederlein: I didn't intentionally make this one, but I'll put it back in.  Not sure how my copy lost that line.

tasks/fix-cat2.yml
```diff
@@ -1,25 +1,30 @@
...
+            src: 01-banner-message.j2
+            dest: /etc/dconf/db/local.d/01-banner-message
+            owner: root
+            group: root
+            mode: '0644'
+        register: rhel_07_010040template
+
+      - name: "MEDIUM | RHEL-07-010030 | RHEL-07-010040 | PATCH | The Red Hat Enterprise Linux operating system must display the Standard Mandatory DoD Notice and Consent Banner before granting local or remote access to the system via a graphical user logon."
+        command: dconf update
+        when: rhel_07_010040template.changed == true
```
```suggestion
        when: rhel_07_010040template is changed
```

dfederlein:
> > > I'm good with using `template` here instead of the `dconf` module. It is a file that contains no other settings other than the two that need to be set by the STIG so it is not likely to stomp on existing configs and it is much simpler/cleaner than dealing with the `dconf` module which is a bit of mess.
> > 
> > 
> > Ack. I guess the issue is that `dconf` module doesn't provide a way to say which db to modify?
> 
> That's one issue, yes. I did realize about the same time you were writing the comments on the dconf command that I needed to conditionalize it. while I didn't use a handler (and I'm open to talking about whether a handler is more appropriate) I did put in a conditional based on the changed status of the previous task.

I will fix this next week: I think you're right it should be a handler.  Let this hang for the weekend and I'll recommit on Monday with a handler instead of the conditional @jamescassell

jamescassell:
> simpler/cleaner than dealing with the `dconf` module which is a bit of mess.

Indeed, you are correct. (Sorry, I hadn't tried; I took the docs at face value.) -- seems it requires python-psutil, which is only included in EPEL, and in a version that's too old at that.  A non-starter for our use.

dfederlein:
> > simpler/cleaner than dealing with the `dconf` module which is a bit of mess.
> 
> Indeed, you are correct. (Sorry, I hadn't tried; I took the docs at face value.) -- seems it requires python-psutil, which is only included in EPEL, and in a version that's too old at that. A non-starter for our use.

Reason #1 to stsick to core modules: community loves epel, and clients don't.

handlers/main.yml
```diff
@@ -83,3 +83,6 @@
 - name: move aide db
   command: "mv {{ rhel7stig_aide_temp_db_file }} {{ rhel7stig_aide_db_file }}"
   when: not rhel7stig_aide_db_status.stat.exists or rhel7stig_overwrite_aide_db
+
+- name: dconfupdate
```
jamescassell: no need to mash them together
```suggestion
- name: dconf update
```

tasks/fix-cat2.yml
```diff
@@ -1,25 +1,21 @@
 ---
-- name: "MEDIUM | RHEL-07-010030 | PATCH | The Red Hat Enterprise Linux operating system must display the Standard Mandatory DoD Notice and Consent Banner before granting local or remote access to the system via a graphical user logon."
-  command: "true"
-  changed_when: no
+### RHEL-07-010030 | RHEL-07-010040 combined as related tasks in regards to a config file no other content will be in.
+- name: "MEDIUM | RHEL-07-010030 | RHEL-07-010040 | PATCH | The Red Hat Enterprise Linux operating system must display the Standard Mandatory DoD Notice and Consent Banner before granting local or remote access to the system via a graphical user logon."
+  template:
```
```suggestion
  copy:
```

tasks/fix-cat2.yml
```diff
@@ -1,25 +1,21 @@
...
+      src: 01-banner-message.j2
```
jamescassell: avoid lots of tiny files; convert literal newlines to "\n"

```suggestion
      content: |
          [org/gnome/login-screen] 
          banner-message-enable=true
          banner-message-text='{{ rhel7stig_logon_banner | replace(newline, "\n") }}'
```

tasks/fix-cat2.yml
```diff
@@ -1,25 +1,21 @@
...
+      dest: /etc/dconf/db/local.d/01-banner-message
+      owner: root
+      group: root
+      mode: '0644'
```
```suggestion
      mode: '0644'
  vars:
      newline: "\n"
```

tasks/fix-cat2.yml
```diff
@@ -40,54 +36,75 @@
       - dod_logon_banner
 
 - name: "MEDIUM | RHEL-07-010060 | PATCH | The Red Hat Enterprise Linux operating system must enable a user session lock until that user re-establishes access using established identification and authentication procedures."
-  command: "true"
-  changed_when: no
+  blockinfile:
+      create: yes
+      dest: /etc/dconf/db/local.d/00-screensaver
+      block: |
+        [org/gnome/desktop/screensaver]
```
jamescassell: yamllint doesn't catch block-strings, but I'd tend to prefer 4 indents here instead of 2; minor nit, though.

tasks/fix-cat2.yml
```diff
@@ -40,54 +36,75 @@
...
       - RHEL-07-010060
-      - notimplemented
 
 - name: "MEDIUM | RHEL-07-010061 | PATCH | The Red Hat Enterprise Linux operating system must uniquely identify and must authenticate users using multifactor authentication via a graphical user logon."
-  command: "true"
-  changed_when: no
+  lineinfile:
+      create: yes
+      dest: /etc/dconf/db/local.d/00-defaults
+      insertafter: '^#[org/gnome/login-screen]'
```
jamescassell: really makes it feel like we should be using `ini_file`

tasks/fix-cat2.yml
```diff
@@ -40,54 +36,75 @@
...
       - RHEL-07-010061
-      - notimplemented
 
 - name: "MEDIUM | RHEL-07-010070 | PATCH | The Red Hat Enterprise Linux operating system must initiate a screensaver after a 15-minute period of inactivity for graphical user interfaces."
-  command: "true"
-  changed_when: no
+  lineinfile:
+      create: yes
+      dest: /etc/dconf/db/local.d/00-screensaver
+      insertafter: '^#[org/gnome/desktop/screensaver]'
```
jamescassell: I don't think you should rely on the presence of `org/gnome/desktop/screensaver` as that's unrelated here
```suggestion
      insertafter: '^#[org/gnome/desktop/session]'
```

tasks/fix-cat2.yml
```diff
@@ -40,54 +36,75 @@
...
+      regexp: '^#?idle-delay'
+      line: idle-delay=uint32 900
+  register: rhel_07_010070linetask
```
jamescassell: No need to register things you aren't going to use.
```suggestion
```

tasks/fix-cat2.yml
```diff
@@ -40,54 +36,75 @@
...
       - RHEL-07-010070
-      - notimplemented
 
 - name: "MEDIUM | RHEL-07-010081 | PATCH | The Red Hat Enterprise Linux operating system must prevent a user from overriding the screensaver lock-delay setting for the graphical user interface."
-  command: "true"
-  changed_when: no
+  lineinfile:
+      create: yes
+      dest: /etc/dconf/db/local.d/locks/session
+      regexp: '^#?lock-delay'
```
jamescassell: which line are you trying to replace?  The `regexp` doesn't match the `line`

tasks/fix-cat2.yml
```diff
@@ -40,54 +36,75 @@
...
       - RHEL-07-010081
-      - notimplemented
 
 - name: "MEDIUM | RHEL-07-010082 | PATCH | The Red Hat Enterprise Linux operating system must prevent a user from overriding the session idle-delay setting for the graphical user interface."
-  command: "true"
-  changed_when: no
+  lineinfile:
+      create: yes
+      dest: /etc/dconf/db/local.d/locks/session
+      regexp: '^#?idle-delay'
```
jamescassell: Which line are you trying to replace?  the `regexp` doesn't match the `line`.

tasks/fix-cat2.yml
```diff
@@ -40,54 +36,75 @@
...
+      line: /org/gnome/desktop/screensaver/idle-delay
+  register: rhel_07_010082
```
jamescassell: you're overwriting the var defined here: https://github.com/MindPointGroup/RHEL7-STIG/blob/3fb4e6a4e5a5b257844f778d57313b048703e850/defaults/main.yml#L78

tasks/fix-cat2.yml
```diff
@@ -98,18 +115,28 @@
...
       - RHEL-07-010100
-      - notimplemented
 
 - name: "MEDIUM | RHEL-07-010101 | PATCH | The Red Hat Enterprise Linux operating system must prevent a user from overriding the screensaver idle-activation-enabled setting for the graphical user interface."
-  command: "true"
-  changed_when: no
+  lineinfile:
+      create: yes
+      dest: /etc/dconf/db/local.d/locks/session
+      regexp: '^#?idle-activation-enabled'
```
jamescassell: `regexp` probably isn't necessary unless needed for cleaning up previous names of this setting, but then a separate task for that might be better.

tasks/fix-cat2.yml
```diff
@@ -98,18 +115,28 @@
...
+      line: /org/gnome/desktop/screensaver/idle-activation-enabled
+  register: rhel_07_010101
```
jamescassell: overwriting https://github.com/MindPointGroup/RHEL7-STIG/blob/3fb4e6a4e5a5b257844f778d57313b048703e850/defaults/main.yml#L82

tasks/fix-cat2.yml
```diff
@@ -118,8 +145,14 @@
       - notimplemented
 
 - name: "MEDIUM | RHEL-07-010110 | PATCH | The Red Hat Enterprise Linux operating system must initiate a session lock for graphical user interfaces when the screensaver is activated."
-  command: "true"
-  changed_when: no
+  lineinfile:
+      create: yes
+      dest: /etc/dconf/db/local.d/00-screensaver
+      insertafter: '^#?[org/gnome/desktop/screensaver]'
```
jamescassell: let's just use `ini_file`

dfederlein: Again, favoring core supported over community supported modules is my goal.

jamescassell: > Again, favoring core supported over community supported modules is my goal.

Let's chat tomorrow at the lockdown meeting https://github.com/ansible/community/wiki/Lockdown#meetings

tasks/fix-cat2.yml
dfederlein: why are we nitpicking this?

tasks/fix-cat2.yml
dfederlein: You are correct, this is a minor nit.

tasks/fix-cat2.yml
dfederlein: This must have been committed pre-cleanup. a recommit will change that on the next round.

tasks/fix-cat2.yml
dfederlein: this is a leftover debugging tool, will be removed.

tasks/fix-cat2.yml
dfederlein: again, leftover debugging, will recommit to clean up.

tasks/fix-cat2.yml
jamescassell: When maintaining the role it's a pain to have to open yet another file to see what's being done, when that file is literally 3 lines.

shepdelacreme: I'm ok with using template over copy here. I actually prefer to have it separated out. It is not that big a deal to reference a template file and if a change needs to be made in the template it limits the commit blast radius to the template file and not the already extremely long `fix-cat2.yml` file. I mean the file is 2500 lines as is, minimizing additional lines does more for readability and maintainability then using copy here.

tasks/fix-cat2.yml
jamescassell: ack on leaving this as `template`.  You still need to replace newlines in the banner with a literal `\n`

tasks/fix-cat2.yml
dfederlein: > Overall it looks good. Thanks for the contribution!
> 
> For these items, the STIG says
> 
> > Note: If the system does not have GNOME installed, this requirement is Not Applicable.
> 
>     * [ ]  We should check and only set these if necessary; at the very least by checking if `dconf` is available, but better by checking for some GNOME package.
> 
>     * [ ]  is there a good reason not to use the `ini_file` module here since that's what we're dealing with?
> 
>     * [ ]  make sure uses of `regexp` make sense
> 
>     * [ ]  remove unnecessary `register` items

For the first, this is handled by the var setting in defaults:

# Whether or not to run tasks related to auditing/patching the desktop environment
rhel7stig_gui: no

Is it not? Or do you want to change this so it is intelligent rather than a knob to turn by the operator running it?

As for the ini module: Yes, I am avoiding community modules for support ability since sometime this year the packaging for community modules and core supported ones will split, and red hat won't support community modules. Since there will be need to support this for MPG customers, and to have red hat support the modules within, avoiding using community modules when possible is paramount.

The register items will need to be recommitted, I need to finish a round of testing before I push that.

PR #250, opened 2019-07-30
touch-create files on vfat efi partition

### LINKED ISSUE #249: Grub task fails on EFI hosts due to vfat fs
Needs ansible PR for a real fix, but can work around by using shell

jamescassell:
ansible tries to set SELinux contexts on these files, which the
underlying fs does not support

fixes #249

related https://github.com/ansible/ansible/pull/59823


jamescassell:
Couple of bugs here, will reopen once they've been fixed.

jamescassell:
reopening.  There was a missing `/` slash character between the directory and the file, and the check_mode ternary was inverted.

jamescassell:
It's unfortunate that this workaround is needed, but the ansible core team doesn't want to backport the proper fix even to ansible 2.8.

PR #251, opened 2019-07-30
allow tmux instead of screen

jamescassell:
auto-detect presence of compliant screen package -- related #239

### DIFF
--- a/tasks/fix-cat2.yml
+++ b/tasks/fix-cat2.yml
@@ -91,7 +91,7 @@
 
 - name: "MEDIUM | RHEL-07-010090 | PATCH | The Red Hat Enterprise Linux operating system must have the screen package installed."
   yum:
-      name: screen
+      name: "{{ rhel7stig_screen_package }}"
       state: present
   when: rhel_07_010090
   tags:
--- a/tasks/fix-cat2.yml
+++ b/tasks/fix-cat2.yml
@@ -89,10 +89,22 @@
       - RHEL-07-010082
       - notimplemented
 
-- name: "MEDIUM | RHEL-07-010090 | PATCH | The Red Hat Enterprise Linux operating system must have the screen package installed."
-  yum:
-      name: "{{ rhel7stig_screen_package }}"
-      state: present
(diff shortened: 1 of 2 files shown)


tasks/fix-cat2.yml
```diff
@@ -89,10 +89,22 @@
...
       - notimplemented
 
-- name: "MEDIUM | RHEL-07-010090 | PATCH | The Red Hat Enterprise Linux operating system must have the screen package installed."
-  yum:
-      name: screen
-      state: present
+- name: "MEDIUM | RHEL-07-010090 | The Red Hat Enterprise Linux operating system must have the screen package installed."
+  block:
+      - name: "MEDIUM | RHEL-07-010090 | AUDIT | The Red Hat Enterprise Linux operating system must have the screen package installed. Check for screen or tmux"
+        command: rpm -q screen tmux
```
shepdelacreme: Do we need this audit check? The `yum` command should be idempotent and I think this rpm check would cause the yum command to skip even if the user decided to change from `screen` to `tmux` as the desired package?

jamescassell: the idea was to be as light-handed as possible in bring the system into compliance. If there is none compliance program installed, don't bother installing another. I could add an additional option if it's desired

shepdelacreme: I think it might be a little unexpected to someone if they set `rhel7stig_screen_package: tmux` but `screen` is already installed and `tmux` doesn't get installed. Or the other way as well.

Maybe it is sufficient to just install whatever they specify in `rhel7stig_screen_package`? I agree we shouldn't remove the other package.

jamescassell: I've renamed the option into a generic variable called `rhel7stig_force_exact_packages` that defaults to the value of `rhel7stig_disruption_high` -- this will cause no extra package to be installed w/o disruption-high, or the exact package to be installed with disruption-high.

(When remediating a system, I generally run with the defaults, followed by disruption_high=yes in check_mode with `--diff` to verify the changes are sane, or otherwise turn off the individual problematic items first.)

PR #260, opened 2019-07-31
quote packages names

jamescassell:
previously this task failed if a failing package name had whitespace


jamescassell:
Assuming close and reopen is the proper way to trigger tests that failed

PR #266, opened 2019-08-02
dconf for GUI

jamescassell:
cleanup of #245

### DIFF
--- a/handlers/main.yml
+++ b/handlers/main.yml
@@ -86,3 +86,4 @@
 
 - name: dconf update
   command: dconf update
+  when: rhel7stig_has_dconf_command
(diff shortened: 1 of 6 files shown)


handlers/main.yml
```diff
@@ -83,3 +83,7 @@
 - name: move aide db
   command: "mv {{ rhel7stig_aide_temp_db_file }} {{ rhel7stig_aide_db_file }}"
   when: not rhel7stig_aide_db_status.stat.exists or rhel7stig_overwrite_aide_db
+
+- name: dconf update
+  command: dconf update
+  when: rhel7stig_dconf_available
```
jamescassell: the same commit that added this also made it invalid, will fix before merging

jamescassell:
Added "disable ctrl-alt-del for gnome" which really shouldn't be a CAT I, except that it is.

PR #284, opened 2019-11-01
add compatability jinja tests

jamescassell:
Ansible assumes these are in place and provided by jinja.

Also grabbed "contains" test from ansible-2.8, "just in case"


jamescassell:
Equivalent Ansible PR: https://github.com/ansible/ansible/pull/64342

PR #287, opened 2019-11-13
enable role for RHEL 8

jamescassell:
These changes make it so that the RHEL7-STIG role can be run cleanly on RHEL 8.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -371,7 +371,7 @@ rhel7stig_force_exact_packages: "{{ rhel7stig_disruption_high }}"
 # RHEL-07-010480 and RHEL-07-010490
 # Password protect the boot loader
 rhel7stig_bootloader_password: 'Boot1tUp!'
-rhel7stig_bootloader_password_hash: "{{ rhel7stig_bootloader_password | grub2_hash(salt=ansible_machine_id) }}"
+rhel7stig_bootloader_password_hash: "{{ 'grub.pbkdf2.sha512.25000.4B656F6B706B4543544A656F44684541355874694C51.9FB9906DFBC1B84A13BA1FCDB2EA6A35E493A6A07ED4506ACF35EAA2F293E24083BFE2FBE4EE842296CC1AEDA26F70E1DD62AC768016DD68FFD8F63C95DFB819' if rhel7stig_bootloader_password == 'Boot1tUp!' else (rhel7stig_bootloader_password | grub2_hash(salt=ansible_machine_id)) }}"
 rhel7stig_boot_superuser: root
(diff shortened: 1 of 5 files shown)


dirtyharrycallahan:
Why do you want to run a RHEL 7 tool against RHEL 8?

jamescassell:
> Why do you want to run a RHEL 7 tool against RHEL 8?

There is no STIG for RHEL 8, but people want to use it and want to harden it to match their existing RHEL 7 deployments in the absence of a STIG for RHEL 8.

dfederlein:
Would this be better split into another repo? (I don't have an opinion either way)

jamescassell:
I've rebased this onto latest devel.  I think we should merge it.

jamescassell:
can we merge this?

defaults/main.yml
```diff
@@ -371,7 +371,7 @@ rhel7stig_force_exact_packages: "{{ rhel7stig_disruption_high }}"
 # RHEL-07-010480 and RHEL-07-010490
 # Password protect the boot loader
 rhel7stig_bootloader_password: 'Boot1tUp!'
-rhel7stig_bootloader_password_hash: "{{ rhel7stig_bootloader_password | grub2_hash(salt=ansible_machine_id) }}"
+rhel7stig_bootloader_password_hash: "{{ 'grub.pbkdf2.sha512.25000.4B656F6B706B4543544A656F44684541355874694C51.9FB9906DFBC1B84A13BA1FCDB2EA6A35E493A6A07ED4506ACF35EAA2F293E24083BFE2FBE4EE842296CC1AEDA26F70E1DD62AC768016DD68FFD8F63C95DFB819' if rhel7stig_bootloader_password == 'Boot1tUp!' else (rhel7stig_bootloader_password | grub2_hash(salt=ansible_machine_id)) }}"
```
dglinder, line 374: Is the hardcoded 'grub.pbkdf2.sha512.[...]' used so the "Boot1tUp!" entry (password hash) is consistent across all systems regardless of the ansible_machine_id?  Why was this chosen over the earlier, more straightforward setting?

jamescassell: It's because there's no python3-passlib package on RHEL 8. I've got a Red Hat case open requesting it be re-added, but no movement yet. If it's useful for you, too, you could open your own case to give it more weight.

dglinder: That makes sense - and if that is the reason that a call to `grub2_hash` won't work on RHEL-8 but would on RHEL-7, then it's a big reason to keep it included.  (Or update Ansible/Jinja2 to not rely on it.)

PR #301, opened 2020-02-05
Changed grub salt to machine_id from ansible_facts.

### LINKED ISSUE #222: grub2_hash uses static salt
We shouldn't hardcode a default salt.  Instead, one should be derived from the inventory_hostname.

Kreeblah:
This is a quick fix to address https://github.com/MindPointGroup/RHEL7-STIG/issues/222.  I've validated that it successfully pulls the machine ID and uses it as the salt in the hash.

fixes #222

### DIFF
--- a/tasks/prelim.yml
+++ b/tasks/prelim.yml
@@ -307,40 +307,11 @@
   when: rhel7stig_ssh_required
 
 - name: "PRELIM | RHEL-07-010482 | RHEL-07-010491 | Gather machine ID"
-  block:
-      - name: "PRELIM | RHEL-07-010482 | RHEL-07-010491 | Check for /etc/machine-id"
-        stat:
-            path: /etc/machine-id
-        register: rhel7stig_etc_machine_id_stat
-
-      - name: "PRELIM | RHEL-07-010482 | RHEL-07-010491 | Gather machine ID from /etc/machine-id"
-        slurp:
-            src: /etc/machine-id
-        register: rhel7stig_raw_etc_machine_id
-        when: rhel7stig_etc_machine_id_stat.stat.exists
-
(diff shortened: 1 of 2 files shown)


Kreeblah:
It turns out that ansible_machine_id doesn't necessarily equate to the contents of /etc/machine-id or /var/lib/dbus/machine-id.  From looking at the Ansible code, I thought that's where it was pulling from, but actually having the prelim steps check those was getting me different results.

So, I updated this to have the prelim steps create a new fact named "rhel7stig_machine_id" that it pulls from whichever of those locations exists, and uses that instead.

Kreeblah:
I must have been SSHed into the wrong host when I thought I was validating the machine ID that was being pulled over.  I didn't realize lookup only reads local files, though, so thanks for catching that.

I updated this to use slurp instead, and made sure I validated the machine ID properly this time.

Kreeblah:
I'm not sure why the build failed, though (maybe a transient error?).  It looks to have failed in an area I didn't touch.

jamescassell:
@Kreeblah I've pushed a commit to your branch; see if that works for you.  @dfederlein note that this will break the grub for RHEL 8 controllers, considering #287

tasks/prelim.yml
```diff
@@ -305,3 +305,18 @@
...
         notify: clean up ssh host key
   when: rhel7stig_ssh_required
+
+- name: "PRELIM | RHEL-07-010482 | RHEL-07-010491 | Gather machine ID"
+  setup:
+      gather_subset: machine_id,!all,!min
+      filter: ansible_machine_id
+  when:
+      - ansible_machine_id is not defined
+  when:
```
jamescassell:
```suggestion
```

PR #304, opened 2020-02-13
Update main.yml informative comment change of appropriated

ocontant:
RHEL-07-040500 is NTP related and not TFTP.

TFTP are covered by RHEL-07-040700 and 040720


jamescassell:
Thanks. I'll verify in a little bit.

PR #310, opened 2020-03-10
Allow UUID home mounts

jeffb4:
Set rhel7stig_using_uuid_fstab to true to mount the /home volume
by its UUID instead of device


jeffb4:
Created to address https://github.com/MindPointGroup/RHEL7-STIG/issues/309

PR #314, opened 2020-03-13
Fix handler playbook output

dglinder:
Remove variable from output to generic "time service" but keep dynamic internal service restart call.

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -1,0 +1,8 @@
+To generate the documentation on a RHEL/CentOS 7 system, take the following steps:
+1. Install required packages:
+   * `yum install python3-pip python-sphinx`
+2. Install the requirements:
+   * `sudo pip3 install -r requirements.txt`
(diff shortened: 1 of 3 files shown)


jamescassell:
could you provide an example of the before/after output? Was it broken before, other than cosmetically?

dglinder:
Sorry @jamescassell  I didn't see your example request.

The fix was cosmetic, but it also simplifies debugging.  Using a variable (e.g. "rhel7stig_time_service ") in the handler name makes it harder to search for an event if you're looking for it quickly.

The old method might show "restart chronyd" on RHEL-7, but "restart ntpd" on RHEL-6 - but neither of those strings are in the playbook code.

The new handler name - "restart time service" - shows consistent output; the end user shouldn't care what actual time service was used, just that it was setup and restarted.  Under the hood the handler restarts the correct service.

PR #315, opened 2020-03-18
[DO NOT MERGE] use ansible-native password_hash filter for grub

jamescassell:


jamescassell:
Was only a proof of concept.  Closing for now.

PR #347, opened 2021-03-18
Fix RHEL-07-010491 grub superusers config

danbarr:
The RHEL-07-010491 task in fix-cat1.yml attempts to check/update `/boot/grub2/grub.cfg`, but this file does not exist on UEFI systems and the task fails. The old version of the role performed this task on `/etc/grub.d/01_users` which seems more proper since direct edits to grub.cfg are overwritten by `grub2-mkconfig`.

Signed-off-by: Dan Barr <dan@dbarr.com>


georgenalen:
Dan,
 Thank you for addressing this issue. We have an automated testing pipeline that does not like PR's from outside of it's own role. I changed the PR to go into staging and then will perform the PR to devel from there. Thanks again, I'll do the same with your other PR

danbarr:
Hi George,
I noticed this and my other fix PR #348 never made it from staging over to devel. Should I re-submit?

georgenalen:
Danbarr,
 Sorry about that we were neck deep in other items and this somewhat fell off. I will work on getting this into devel today.

George

georgenalen:
Danbarr, 
 As an update I merged our changes into devel this morning. It was a bit out of sync with some other devel pushes so I had to get that sorted, but all is good. It will move into main with the May fixes push. 

George

danbarr:
Awesome, thanks!

PR #359, opened 2021-06-16
Fix RHEL-07-010050 default enable/disable setting

danbarr:
The /etc/issue and /etc/issue.net banners shouldn't be dependent on GUI

Signed-off-by: Dan Barr <dan@dbarr.com>


uk-bolly:
Hi @danbarr 

Thank you again for raising issues when you find them and providing the solution. We are still working to get the pipeline setup so all PRs are enabled to check by default.
To save waiting any longer for this to be put in place i have pulled yours into #361 (hopefully approved and pulled into devel later today). so that we can get you commit added and checked in the pipeline.

Thank you once again for continuing to assisting us in improving the roles.

uk-bolly

PR #362, opened 2021-07-02
Fix for too much info in perms_results var

georgenalen:
Signed-off-by: George Nalen <georgen@mindpointgroup.com>


uk-bolly (approved):
looks good and tested

Issue #351, opened 2021-04-01
/etc/login.defs CREATE_HOME enabled value should be "yes" instead of "True"


waweber:
RHEL-07-020610 - "The Red Hat Enterprise Linux operating system must be configured so that all local interactive user accounts, upon creation, are assigned a home directory." checks this state by ensuring '^\s*CREATE_HOME\s+(\S+)\s*$' has the subexpression equal to "yes", but defaults/main.yml has it set to "true". Not sure if this is some sort of Ansible boolean parsing magic (version 2.10.5)

The following diff should fix it:

```
diff --git a/defaults/main.yml b/defaults/main.yml
index dd09cf0..fc52811 100644
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -599,7 +599,7 @@ rhel7stig_login_defaults:
     pass_max_days: 60
     fail_delay_secs: 4
     umask: '077'
-    create_home: 'true'
+    create_home: 'yes'
 
 # RHEL-07-030300 uncomment and set the value to a remote IP address that can receive audit logs
 # rhel7stig_audisp_remote_server: 10.10.10.10
```

georgenalen:
I fixed this with PR #352 that is in devel right now. Once we get a few other updates to the role I will role it in with the a new release version and merge into main. Thanks for finding this one!

George

georgenalen:
Addressed with PR #364

PR #364, opened 2021-07-06
Issue Fixes and Enhancements

georgenalen:
Issue Fixes:
#351 - login.defs values should be yes instead of true
#362 - Disable USB Storage could be done differently 

Enhancements:
- linting and syntax
- Addressed too restrictive perms in perms_results variable
- Addressed EFI audit check issue discovered
- Updated goss audit info and vars
- Address comments regarding access reqs
- fixed typos
- addressed default var value for rhel_07_010050

Signed-off-by: George Nalen <georgen@mindpointgroup.com>


uk-bolly (approved):
tried and tested

PR #411, opened 2022-06-09
Fix rhel7stig_run_audit var

bordenit:
rhel7stig_run_audit is mentioned in the README, but all the when conditions say run_audit. Var should be updated to reflect the README.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -28,7 +28,7 @@ get_audit_binary_method: download
 audit_content: git
 
 # enable audits to run - this  runs the audit and get the latest content
-run_audit: false
+rhel7stig_run_audit: false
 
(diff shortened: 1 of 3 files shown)


uk-bolly:
hi @bordenit 

Thats a great catch, thank you for taking the time to raise this issue and provide a fix. In this instance the README is incorrect we are moving to a common variable for the audit portion across all the OS's. 
There is a new update to the audit script itself coming also.

I am happy to change it unless you want to in this PR and we can merge this with the README fix?

Many thanks once again

uk-bolly

bordenit:
@uk-bolly Thanks you can go ahead and make the change. Also, looks like the test might be failing due to extra_var not being passed in for: rhel7stig_bootloader_password_hash: 'grub.pbkdf2.sha512.thishasbeenchangedfortesting'

Thanks

PR #425, opened 2023-04-26
Remove use of deprecated shell module warnings feature in Ansible

ericamador:
Currently, this role can break when used with Ansible 2.14 and later because of the use of the warn argument to suppress warning output has been removed. Ansible no longer displays these warnings as of version 2.11. These change simply remove the use of this argument when using the shell module in order to restore compatibility with newer versions of Ansible.

See this PR for more information on the deprecated warnings feature in Ansible: https://github.com/ansible/ansible/pull/70504

N/A

### DIFF
--- a/handlers/main.yml
+++ b/handlers/main.yml
@@ -58,8 +58,6 @@
 
 - name: restart auditd
   ansible.builtin.shell: /usr/sbin/service "{{ rhel7stig_audit_daemon }}" restart
-  args:
-      warn: false
 
 - name: rebuild initramfs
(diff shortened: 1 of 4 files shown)


uk-bolly:
hi @ericamador 

Thank you so much for the PR, this looks great. It is however just missing both gpg key and signed-off requirements. If you could updated we will be happy to accept this PR.

regards

uk-bolly

ericamador:
@uk-bolly Thanks, I have completed those gpg key and sign-off requirements.

uk-bolly (approved):
Great find

PR #429, opened 2023-07-26
Improve sudo user password check

uk-bolly:
Overall Review of Changes:
Improve password check for rule RHEL-07-010340 that sudo user must have a password or sudo will fail to work.

Issue Fixes:
https://github.com/ansible-lockdown/RHEL7-STIG/issues/428

Enhancements:
Only run if the playbook is not being run as a super user

How has this been tested?:
Manually


MrSteve81 (approved):
Looks Good!

PR #430, opened 2023-08-03
Update fix-cat2.yml

whitehat237:
Removes quote  `"` character from module parameter.  This causes a task failure and ansible reports that it's an unknown module parameter.

Modifies lines 155, and 166 and removes extraneous quote character

**Issue Fixes:**
I have not opened an issue for this but can do so if needed.
Resolves an issue in fix-cat2.yml with the task for RHEL-07-010063
Ansible reports that create\" is an unsupported module parameter for lineinfile module.

**Enhancements:**
None

Made the modifications describe above, and ran the stig role against a RHEL 7.9 system and failed task RHEL-07-010063 now works as expected.

### DIFF
--- a/tasks/fix-cat2.yml
+++ b/tasks/fix-cat2.yml
@@ -1350,7 +1350,7 @@
             enabled: false
             state: stopped
         when:
-            - rhel_07_020110_autofs_service_status == "loaded"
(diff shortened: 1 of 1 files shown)


uk-bolly (approved):
Brilliant find

PR #442, opened 2023-10-31
updated the workflow version and galaxy setup

uk-bolly:
Update to workflows

move action/checkout to v4
use the sleep variable for sleep time during run (some OSs going slower on build)
update the update_galaxy role to use new code for galaxy-ng


MrSteve81 (approved):
Approved

PR #445, opened 2023-11-14
rhel7stig_boot_part variable now discovered

uk-bolly:
rule_021350 when set false caused failures.
Created discovery for system and default var removed

manually on bios and efi based systems


georgenalen (approved):
looks good

Issue #462, opened 2024-04-24
RHEL-07-020620 uidmin and uidmax not sufficient to find local interactive users


prestonSeaman2:
**Describe the Issue**
Using uidmin and uidmax is not sufficient to find local interactive users.  Some users with sbin nologin are getting set as local interactive users when they are not.

**Expected Behavior**
Users with login shells set to sbin nologin should not be placed in the list of local interactive users.

**Actual Behavior**
Users with login shells set to sbin nologin are sometimes placed in the list of local interactive users.

**Control(s) Affected**
RHEL-07-020620 and any control/task using the rhel7_passwd fact.

**Possible Solution**
Create new fact which is a subset of the original rhel7_passwd which has local interactive users stored in it.  The fact would be populated using logic checking 'uid' >= rhel7_stig_interactive_uid_start | int, uid <= rhel7_stig_interactive_uid_stop | int uid != 65534, shell != /sbin/nologin

layluke:
I've ran into the same issue lately, This bug kept on exhibiting itself by causing some issues with nfs. Turns out since /var/lib/nfs is set as the home directory for nfsnobody and rpcuser, it was locking the directory to 700 permissions, which caused some issues with rpcuser reading a lock file in the directory.

> uidmin and uidmax sometimes set to empty string. In these cases the login shell is the only way to check if they are local interactive users.
Yeah,  best way to handle this I think would be to exclude /usr/sbin/nologin and /sbin/nologin with grep -v when /etc/passwd is parsed.

uk-bolly:
hi @layluke and @prestonSeaman2 

Thank you again for this issue. The fix has been applied and is now in the main branch and release. I will close this issue, Please reopen if you feel this is not resolved.

Many thanks

uk-bolly

PR #466, opened 2024-05-02
Excluding non-interactive logins shells from being parsed

layluke:
Overall Review of Changes:
Excludes non interactive shells listed in /etc/shells from being registered in the rhel7stig_passwd_file_audit variable

Issue Fixes:
https://github.com/ansible-lockdown/RHEL7-STIG/issues/462

Enhancements:
N/A

How has this been tested?:
tested code in separate task, results were as expected


uk-bolly (approved):
nice improvement

